PR #20, opened 2017-01-05
Clean up readme

PavelSynek:
I think BarCode should be Barcode instead, but since the library is using BarCode class, I used that.


alexio:
@PavelSynek thanks for the contribution :D

PR #21, opened 2017-01-05
No getValue(). Did you mean getKey()?

PaulWoitaschek:
In BarCode


digitalbuddha:
We sure did! thanks for finding/fixing

PR #24, opened 2017-01-06
Update/Clean up build.gradle

eneim:
Remove one "empty line", remove duplicated dependency entry (Google gsm).


build.gradle
```diff
@@ -24,7 +23,6 @@ buildscript {
         classpath 'net.ltgt.gradle:gradle-apt-plugin:0.5' // for non-android projects
```
digitalbuddha: I don't believe we are using this apt either.  Want to give it a yank and see if it still builds?

eneim: @digitalbuddha There is a comment about "non-android projects" so I didn't touch. But let me try clean it and push to see if it still builds.

digitalbuddha:
Thank you for contribution

PR #33, opened 2017-01-08
Fix typo : Change "us" to "use"

anupcowkur:


digitalbuddha:
Thanks for contributing!

Issue #30, opened 2017-01-07
Create Factory For SourcePersister & Middleware


digitalbuddha:
It would be useful to have factories rather than calling the implementations directly for:
`SourcePersister`
`GsonReaderParse`
`GsonSourceParse`
`GsonStringParser`

scottmeschke:
Opened up a PR for this.

PR #35, opened 2017-01-08
Added factories for SourcePersisters and the various GsonParser imple…

scottmeschke:
"Added factories for SourcePersisters and the various GsonParser implementations. Updated README to use new factories. Updated example activity to use new factories. All factories return the general Store interfaces."

### DIFF
--- a/filesystem/src/main/java/com/nytimes/android/external/fs/SourcePersisterFactory.java
+++ b/filesystem/src/main/java/com/nytimes/android/external/fs/SourcePersisterFactory.java
@@ -9,7 +9,7 @@ import java.io.IOException;
 import okio.BufferedSource;
 
 /**
- * Created by Scott on 1/8/2017.
+ * Factory for {@link SourcePersister}
  */
 
 public class SourcePersisterFactory {
@@ -17,8 +17,6 @@ public class SourcePersisterFactory {
     /**
      * Returns a new {@link BufferedSource} persister with the provided file as the root of the
      * persistence {@link com.nytimes.android.external.fs.filesystem.FileSystem}.
-     * @param root
-     * @return
      * @throws IOException
      */
     public static Persister<BufferedSource> create(File root) throws IOException {
--- a/filesystem/src/main/java/com/nytimes/android/external/fs/SourcePersisterFactory.java
+++ b/filesystem/src/main/java/com/nytimes/android/external/fs/SourcePersisterFactory.java
@@ -1,0 +1,28 @@
+package com.nytimes.android.external.fs;
+
+import com.nytimes.android.external.fs.filesystem.FileSystemFactory;
+import com.nytimes.android.external.store.base.Persister;
+
+import java.io.File;
+import java.io.IOException;
+
+import okio.BufferedSource;
+
+/**
+ * Created by Scott on 1/8/2017.
+ */
+
+public class SourcePersisterFactory {
+
+    /**
+     * Returns a new {@link BufferedSource} persister with the provided file as the root of the
+     * persistence {@link com.nytimes.android.external.fs.filesystem.FileSystem}.
+     * @param root
+     * @return
+     * @throws IOException
+     */
+    public static Persister<BufferedSource> create(File root) throws IOException {
+        if (root == null) throw new IllegalArgumentException("root file cannot be null.");
+        return new SourcePersister(FileSystemFactory.create(root));
+    }
+}
--- a/middleware/src/main/java/com/nytimes/android/external/store/middleware/GsonParserFactory.java
+++ b/middleware/src/main/java/com/nytimes/android/external/store/middleware/GsonParserFactory.java
@@ -1,0 +1,73 @@
+package com.nytimes.android.external.store.middleware;
+
+import com.google.gson.Gson;
+import com.nytimes.android.external.store.base.Parser;
+
+import java.io.Reader;
+
+import okio.BufferedSource;
+
+/**
+ * Factory which returns various Gson {@link Parser} implementations.
+ */
+
+public class GsonParserFactory {
+
+    /**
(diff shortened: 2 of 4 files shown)


digitalbuddha:
Thanks for contribution! Closes https://github.com/NYTimes/Store/issues/30

middleware/src/main/java/com/nytimes/android/external/store/middleware/GsonParserFactory.java
```diff
@@ -0,0 +1,73 @@
...
+        return new GsonReaderParser<>(gson,parsedClass);
+    }
+
+    /**
+     * Returns a new Parser which parses from {@link Reader} to the specified type, using
+     * a new default configured {@link Gson} instance.
+     */
+    public static <T> Parser<Reader, T> createReaderParser(Class<T> parsedClass) {
+        if (parsedClass == null) throw new IllegalArgumentException("parsedClass cannot be null.");
+        return new GsonReaderParser<>(new Gson(), parsedClass);
```
digitalbuddha: This is nice. Great addition!

digitalbuddha: Maybe the gson instance should be a singleton. My thinking is why create more than one if multiple stores want to use the convenience method.

scottmeschke: Yeah I had the same idea. 
Not sure if that's something you would want to provide in the library, or have them write themselves.

scottmeschke:
Also, @digitalbuddha wasn't sure what the opinions on Precondition assertions were, as I didn't see any in the code besides the transcribed Guava cache code. 

That being said, my personal preference is that I'm **always** happy when a library code fails immediately and clearly (of course at the cost of some boilerplate in the library code).

middleware/src/main/java/com/nytimes/android/external/store/middleware/GsonParserFactory.java
digitalbuddha: @brianPlummer @alexio thoughts?

scottmeschke: The other trick is that the factory is all static and has no state. 
I'm not a fan of static variables that are mutable.

digitalbuddha:
I agree I like precondition checks as well. Feel free to add

digitalbuddha:
Let's wait for 1 more collab review before merging but lgtm. I agree on not holding static state.

michaldrabik:
I was also thinking about missing Preconditions when doing #28, but decided to not add any because there wasn't any in the codebase. 

Maybe we could create an issue for that so a single PR can be made to add preconditions for the whole library (where necessary).

digitalbuddha:
Sure ticket made https://github.com/NYTimes/Store/issues/41

middleware/src/main/java/com/nytimes/android/external/store/middleware/GsonParserFactory.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Returns a new Parser which parses from {@link Reader} to the specified type, using
+     * a new default configured {@link Gson} instance.
+     */
+    public static <T> Parser<BufferedSource, T> createSourceParser(Class<T> parsedClass) {
+        if (parsedClass == null) throw new IllegalArgumentException("parsedClass cannot be null.");
+        return new GsonSourceParser<>(new Gson(), parsedClass);
+    }
+
+    /**
```
digitalbuddha: Comment here and on next one seem to be reversed

middleware/src/main/java/com/nytimes/android/external/store/middleware/GsonParserFactory.java
alexio: I would vote against maintaining a singleton gson instance. If multiple instances are an issue, caller can resolve it by passing in their own.

alexio: I would vote against maintaining a singleton, if multiple instances are an issue the caller can resolve by passing in its own.

filesystem/src/main/java/com/nytimes/android/external/fs/SourcePersisterFactory.java
```diff
@@ -0,0 +1,28 @@
...
+import com.nytimes.android.external.fs.filesystem.FileSystemFactory;
+import com.nytimes.android.external.store.base.Persister;
+
+import java.io.File;
+import java.io.IOException;
+
+import okio.BufferedSource;
+
+/**
+ * Created by Scott on 1/8/2017.
```
digitalbuddha: actually just remove it, we can go to github to see who made file.

digitalbuddha:
fix the top level comment & flip the other two and this should be good to go.

brianPlummer:
@scottmeschke looks good! strip out the "created by" at the top of SourcePersisterFactory.java and it should be good to merge in!! :) thank you so much!!

filesystem/src/main/java/com/nytimes/android/external/fs/SourcePersisterFactory.java
scottmeschke: did mean to keep the "created" thing, just IDE setting I forgot to change, I'll fix.

scottmeschke:
at work now, I'll fix these issues at home tonight. ty for the review

scottmeschke:
Fixed the javadoc mistakes.

digitalbuddha:
Thank you do much for the contribution! We haven't even setup checkstyle yet. Love this community 💯

PR #44, opened 2017-01-10
Adds middleware for top-level json array parsing

Jogan:


middleware/src/main/java/com/nytimes/android/external/store/middleware/GsonSourceListParser.java
```diff
@@ -0,0 +1,41 @@
...
+ * example usage:
+ * ParsingStoreBuilder.<BufferedSource, List<BookResults>>builder()
+ * .fetcher(fetcher)
+ * .persister(new SourcePersister(fileSystem))
+ * .parser(new GsonSourceListParser<>(gson, new TypeToken<List<BookResults>())
+ * .open();
+ */
+
+
+public class GsonSourceListParser<Parsed> implements Parser<BufferedSource, Parsed> {
```
digitalbuddha: Could you kindly add a unit test validating that this parser works with List<T>

digitalbuddha: I also think the example in comment should be using the factory.

digitalbuddha:
Thank you for contribution!

PR #48, opened 2017-01-11
start of checkstyle inclusion...

brianPlummer:
* an initial pr to get the work going on checkstyle
* this fails the build....should we turn that off? merge it in and fix stuff in a different pr?


digitalbuddha:
Change Travis.yml.to not run check for now. Have it run just test.

Issue #39, opened 2017-01-09
Create Moshi & Jackson Middleware 


digitalbuddha:
Following same structure as current Middleware it would be nice to also have 2 additional modules: middleware-jackson and Middleware-moshi.  We will add this eventually and welcome an external contribution.

PR #55, opened 2017-01-14
Create Moshi Middleware

maksim-m:
Related: #39


middleware-moshi/src/main/java/com/nytimes/android/external/store/middleware/moshi/MoshiReaderParser.java
```diff
@@ -0,0 +1,39 @@
...
+    public MoshiReaderParser(@NonNull Moshi moshi, @NonNull Type type) {
+        jsonAdapter = moshi.adapter(type);
+    }
+
+    @Override
+    @Nullable
+    public Parsed call(@NonNull Reader source) {
+        String jsonString;
+        try {
+            jsonString = IOUtils.toString(source);
```
digitalbuddha: I'd prefer to not have apache commons as additional dependency. Also why convert to String rather than  feeding reading into moshi. I'm not too familiar with moshi but I'm sure it has a streaming api right?

digitalbuddha: If not from this one. Also check maybe you can go from reader to Buffered source

digitalbuddha:
Looked at it more. There's no reason to do Reader as an input to moshi. Remove that one, keep other two and we'll wait for one more reviewer. Thanks for contributing

maksim-m:
@digitalbuddha You're right, thanks for the feedback. Fixed.

digitalbuddha:
I'll do a release Monday..

PR #56, opened 2017-01-14
Preconditions for middleware module

michaldrabik:
This PR introduces preconditions for `middleware` module. Using `Preconditions` class from already present Guava. Also refactored factory to use overloading and added tests.


digitalbuddha:
Lgtm nice refactor.

Issue #57, opened 2017-01-14
Allow multiple parsers. 


digitalbuddha:
Currently you can only pass in a single parser. There are times when I want to use a Middleware parse plus add additional parser. 

Ideally I should be able to do:
Builder.fetcher().parser(firstParser).parser(secondParser).build()

A solution should be found that does not require any additional generics while allowing any number of parsers to be used in serial.

michaldrabik:
What would a real life scenario be, where you need multiple parsers?

digitalbuddha:
Let's say when you first want to use a Middleware parser to convert from json to pojos. Then you want to rearrange the data for example flattening an api that has nested nodes

michaldrabik:
OK, so like: JSON --> POJO1 --> POJO2 (different type than POJO1)

digitalbuddha:
Yup! It will make it easier to share some parsers and then build on top of them without needing to subclass

PR #61, opened 2017-01-15
Multi Parser

michaldrabik:
This PR is a proposal of resolving #57.

I'm not completely sure if this is what we are looking for and if this can be done in a prettier way. 
Anyway, leaving here for review, discussion.

### DIFF
--- a/store/src/main/java/com/nytimes/android/external/store/base/impl/ParsingStoreBuilder.java
+++ b/store/src/main/java/com/nytimes/android/external/store/base/impl/ParsingStoreBuilder.java
@@ -11,6 +11,8 @@ import com.nytimes.android.external.store.base.Persister;
 import com.nytimes.android.external.store.base.Store;
 import com.nytimes.android.external.store.util.NoopPersister;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.Callable;
 
 import rx.Observable;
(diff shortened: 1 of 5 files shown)


store/src/main/java/com/nytimes/android/external/store/base/impl/ParsingStoreBuilder.java
```diff
@@ -81,13 +84,22 @@ public Raw call() throws Exception {
...
     @NonNull
     public ParsingStoreBuilder<Raw, Parsed> parser(final @NonNull Func1<Raw, Parsed> parser) {
-        this.parser = parser;
+        this.parsers.clear();
+        this.parsers.add((Parser<Raw, Parsed>) parser);
         return this;
     }
 
     @NonNull
     public ParsingStoreBuilder<Raw, Parsed> parser(final @NonNull Parser<Raw, Parsed> parser) {
```
digitalbuddha: Any reason to clear when adding another parser? Not saying it should exist just trying to figure out logic behind not allowing someone to call builder.parser().parser().build() and expecting 2 parsers to be added?

michaldrabik:
```
        Store<String> simpleStore = ParsingStoreBuilder.<String, String>builder()
                .persister(persister)
                .fetcher(fetcher)
                .parser(parser1)
                .parser(parser2)
                .open();
```

In the above example we want <String, String> Store so parser1 would have to be <String, X> and parser2 <X, String>. It would be impossible to achive with current generics.

We would need to remove generics from arguments: 

```
    @NonNull
    public ParsingStoreBuilder<Raw, Parsed> parser(final @NonNull Func1 parser) {
        this.parsers.add((Parser) parser);
        return this;
    }

    @NonNull
    public ParsingStoreBuilder<Raw, Parsed> parser(final @NonNull Parser parser) {
        this.parsers.add(parser);
        return this;
    }

    @NonNull
    public ParsingStoreBuilder<Raw, Parsed> parsers(final @NonNull List<Parser> parsers) {
        this.parsers.addAll(parsers);
        return this;
    }
```

brianPlummer: i'm fine with it clearing the list..

Issue #23, opened 2017-01-05
Change Request: Consider migrating Barcode to a Barcode<T>


PaulWoitaschek:
I find the BarCode class very limiting, because:

* A String,String key value type rarely matches my key type
* Its not type safe. If you just have a single int as a key you need to remember to set it as the string value of the key and follow that convention. When you need for example an int, the `Integer.parseInt()` makes it even more dangerous

I suggest to replace the BarCode class completely with generics.

Then the sample would just look like:
```java
  private Store<RedditData, String> provideRedditStore() {
    return StoreBuilder.<RedditData, String>builder()
      .fetcher(key -> provideRetrofit().fetchSubreddit(key, "10"))
      .open();
  }

  public void loadPosts() {
    String awwRequest = "aww";
    provideRedditStore()
      .get(awwRequest)
      .flatMap(this::sanitizeData)
      .toList()
      .subscribeOn(Schedulers.io())
      .observeOn(AndroidSchedulers.mainThread())
      .subscribe(this::showPosts, throwable -> {
        Log.e(StoreActivity.class.getSimpleName(), throwable.getMessage(), throwable);
      });
  }
```

digitalbuddha:
Thanks for the suggestion. We actually started down that route and backed off from it.  For one thing the ParsingStoreBuilder used to need 3 generics `<Raw,Parsed,Barcode>`.  Additionally all the user implemented pieces need to carry an additional generic for the barcode.  The thing is that this still doesn't do any type of auto mapping (for example, if your barcode was RequestPOJO, you still need to know which fields need to be read out in the `fetcher` & `persister` and read them in/out yourself).

Internally, we go with a solution of using a gson serializer for any keys that need to contain more than 2 values or want to be differently named.  We then can do something like `Barcode=BarcodeCreator.from(field1,field2,field3)` when creating the barcode and `BarcodeCreator.decode` when using it within each layer to turn back into a particular pojo.

My rational was that try to keep this as simple as possible.  

In theory there's nothing stopping you from creating `MyBarcode<T> extends Barcode` and then casting to the subclass when you need it to read your values. (Just remember that the Barcode is also used as a key within the in-memory cache).

Thank you for suggestion, I'll bring it up at next internal dev meeting and see how team feels about it.

PaulWoitaschek:
Any news on this?

I tried your approach:
[...]

PR #64, opened 2017-01-16
Experimental Store Code Gen Part 1

digitalbuddha:


digitalbuddha:
Store-Base not building due to being a java module and depending on android annotations.  Need to change to some other nullability annotations.  Open to ideas of how else to organize the project. Currently anything that is needed by the annotation processor has been moved to a Store-Base java only module.  

This PR contains code gen for `BarCode` that are specific to a retrofit @Get method.  Next step is cleanup & code gen to create a StoreModule which has a `provides` method for each Store.  This way you can auto generate stores and pass in the file system and gson instances (see example in the comment below `Api.java`

digitalbuddha:
This PR should allow a fix for #23 without carrying through another generic

digitalbuddha:
from ```
@BuildStore
public interface Api {

    @GET("r/{subredditName}/new/.json")
    @Persister
    Observable<RedditData> fetchSubreddit(@Path("subredditName") String subredditName,
                                          @Query("limit") String limit);
```

We can generate
```
@BuildStore
public interface Api {

    @GET("r/{subredditName}/new/.json")
    @Persister
    Observable<RedditData> fetchSubreddit(@Path("subredditName") String subredditName,
                                          @Query("limit") String limit);
```
Obviously some cleanup needed but we are able to generate specific barcode and provides methods that generate each store.

digitalbuddha:
close this for now. needs a bit more work before it can be landed. For one thing barcode generation needs to generate proper hashcode and equals

PR #70, opened 2017-01-19
Fix memory caching issue with Equivalence.Equals

shunyy:
Currently, memory cache can not handle multiple `BarCode` instance that have same contents as expected. Bud disk cache can handle that as expected.

This PR implements `BarCode.hashCode` and `BarCode.equals` so that `Equivalence.Equals` works as expected.


digitalbuddha:
Thank you for contribution. Somehow was missed when we moved this external.

Issue #72, opened 2017-01-20
Clarifying Store.stream


PaulWoitaschek:
Currently I find the implementation of `stream(barcode)` very confusing. 

If I understand correctly, it makes a call with the barcode and then returns all results of all barcodes.
But before reading the documentation I thought that it would stream everything that matches to the barcode.

My suggestion is to remove the barcode from the signature of stream. And to make stream() completely passive (like the method implicates to me).

If someone wants the behavior of what it does currently, he could just use RxJava to archieve exactly that, i.e. `Observable.conat(store.get(barcode), store.stream())`.

digitalbuddha:
Makes sense. Want to PR?

PaulWoitaschek:
I'll give it a try.

digitalbuddha:
Sure. I can get it done over weekend too. Deprecated current implementation, I'll pull it out in 1.1

PaulWoitaschek:
Like this? Couldn't resist from making a small cleanup.

PR #73, opened 2017-01-20
Added stream() for #72

PaulWoitaschek:
* Added the stream method
* Deprecated old stream(barcode) and added migration helping in the documentation
* Added missing Nullable/NonNullable annotations
* Added missing Override annotations
* Fixed futures generic hickup

### DIFF
--- a/cache/src/main/java/com/nytimes/android/external/cache/Futures.java
+++ b/cache/src/main/java/com/nytimes/android/external/cache/Futures.java
@@ -21,10 +21,10 @@ public final class Futures {
     @Nullable
     public static <V> ListenableFuture<V> immediateFuture(@Nullable V value) {
         if (value == null) {
-            Futures.ImmediateSuccessfulFuture typedNull = Futures.ImmediateSuccessfulFuture.NULL;
-            return typedNull;
+            //noinspection unchecked safe because of erasure
+            return (ListenableFuture<V>) ImmediateSuccessfulFuture.NULL;
         } else {
-            return new Futures.ImmediateSuccessfulFuture(value);
+            return new Futures.ImmediateSuccessfulFuture<>(value);
         }
     }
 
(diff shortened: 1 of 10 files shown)


cache/src/main/java/com/nytimes/android/external/cache/Futures.java
```diff
@@ -21,24 +21,24 @@ private Futures() {
     @Nullable
     public static <V> ListenableFuture<V> immediateFuture(@Nullable V value) {
         if (value == null) {
-            Futures.ImmediateSuccessfulFuture typedNull = Futures.ImmediateSuccessfulFuture.NULL;
-            return typedNull;
+            //noinspection unchecked safe because of erasure
+            return (ListenableFuture<V>) ImmediateSuccessfulFuture.NULL;
         } else {
-            return new Futures.ImmediateSuccessfulFuture(value);
+            return new Futures.ImmediateSuccessfulFuture<>(value);
```
digitalbuddha: lol you're fixing guava code nice!

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -227,6 +231,11 @@ void notifySubscribers(Parsed data) {
         return stream;
     }
 
+    @NonNull @Override
+    public Observable<Parsed> stream() {
+        return subject.asObservable();
```
digitalbuddha: lets add a `.share` before giving this out, should have been there all along 
http://blog.danlew.net/2016/06/13/multicasting-in-rxjava/

PaulWoitaschek: Why? It's a subject and it can have multiple subscribers.

digitalbuddha: Observable<Foo> fooObservable = store.stream();

barObservable= fooObservable.map(foo->foo.bar());

barObservable.subscribe()
barObservable.subscribe()

map gets called twice.  I'd prefer for it to be called once

digitalbuddha: actually forget what I said, this is different problem

Issue #68, opened 2017-01-18
Repeat When Clear functionality


digitalbuddha:
Update/EDIT:
It would be nice to have a repeat when functionality where anytime clear is called a subscriber will be resubscribed.  The new public api method will be `store.getRefreshing(barcode)` a call to store.clear() drives the refresh

We also need to add clearAll(barcode) which will clear both memory and disk cache.

digitalbuddha:
Might as well add isExpired if adding isStale to instead skip the disk cache completely

PaulWoitaschek:
What I need is also way to directly remove entries belonging to a specific barcode.

From my perspective what you call cache is only an implementation detail for me. When I call `clearMemory` what I really want is to clear both the memory cache as well as the file persistence cache. 

My usecase is to be consistent to the backend.

I have an endpoint A that I call with a date that gives me the list of trainings.
I have an endpoint B where I can post a training for a date.

When I now add a Training in B, I directly want to tell the library: Entries for this date are invalid. Remove them immediately so the next call to A will include the just posted training.

What I want this library to do for me is to expose a stream of data that is the source of truth.
When I invalidate a cache entry, I send the barcode (or the date) to a `PublishSubject`.
After each `store.get(barcode)` I apply a `repeatWhen(cacheEntrySubject....)` emits that barcode so the actual stream never stops and I can subscribe from any place I want as long as I make sure I invalidate the entries when pushing new ones to the backend.

digitalbuddha:
Thats an interesting way to approach  a solution. Currently what I do I  your case is call store.fetch to refill the cache with new entry so that next call to get will have the new entry.   Let me think of a way to implement your requirement

pavlospt:
@PaulWoitaschek Why would you hit the cache to request data when you are sure that there are dates that should have been invalidated? I mean I would just go with `fetch` to be in sync with backend and then call `get` on the rest of the app.

PaulWoitaschek:
I need to access each endpoint from about 2-5 places at the same time. Each place is completely agnostic of the others. Based upon user interaction all these places need to refresh data. So there is no first place where I can call `fetch` from.

digitalbuddha:
If I understand correctly you want a subject which is endless and emits a barcode any time you call clear with a barcode? What do you want it to emit when  there is a call for clear all?

pavlospt:
[...]

PR #86, opened 2017-01-29
Migrate Barcode to any Type

### LINKED ISSUE #23: Change Request: Consider migrating Barcode to a Barcode<T>

digitalbuddha:
close #23 

This is a breaking change. It will be open a bit for discussion. This change introduces a new StoreBuilder architecture. Rather than having a `StoreBuilder` and a `ParsingStoreBuilder` with different generics (1 vs 2).  We now have a single StoreBuilder that can take in 3 types:  Raw, Parsed, Key (BarCode).  Api looks like:
```java
Store<Date, Integer> store = StoreBuilder
               .fromTypes(Key.class, Parsed.class, Raw.class)
               .fetcher()
               .parser()
               .persister()
```

This allows us to support from 1 to 3 types being passed in.  The problem with this apporach is that we need to change the signature of fetcher<Raw> to instead be fetcher<Raw,Barcode> to support a custom `Barcode` type to be passed in, similar with the `Persister` and the generic supporting `Store` as seen above.

Opening this PR for discussion. External opinions are greatly appreciated

### DIFF
--- a/store/src/main/java/com/nytimes/android/external/store/base/impl/ParsingStoreBuilder.java
+++ b/store/src/main/java/com/nytimes/android/external/store/base/impl/ParsingStoreBuilder.java
@@ -9,11 +9,11 @@ import com.nytimes.android.external.store.base.Persister;
 import com.nytimes.android.external.store.base.Store;
 import com.nytimes.android.external.store.util.NoopPersister;
 
-import org.jetbrains.annotations.Nonnull;
-
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Callable;
+
+import javax.annotation.Nonnull;
 
 import rx.Observable;
 import rx.functions.Func1;
@@ -84,13 +84,6 @@ public class ParsingStoreBuilder<Raw, Parsed> {
                 return diskWrite.write(barCode, raw);
             }
         };
-        return this;
-    }
-
-    @Nonnull
-    public ParsingStoreBuilder<Raw, Parsed> parser(final @Nonnull Func1<Raw, Parsed> parser) {
-        this.parsers.clear();
-        this.parsers.add((Parser<Raw, Parsed>) parser);
         return this;
     }
 
--- a/store/src/main/java/com/nytimes/android/external/store/base/impl/ParsingStoreBuilder.java
+++ b/store/src/main/java/com/nytimes/android/external/store/base/impl/ParsingStoreBuilder.java
@@ -95,7 +95,7 @@ public class ParsingStoreBuilder<Raw, Parsed> {
     }
 
     @Nonnull
-    public ParsingStoreBuilder<Raw, Parsed> parsers(final @Nonnull List<Parser<Raw, Parsed>> parsers) {
+    public ParsingStoreBuilder<Raw, Parsed> parsers(final @Nonnull List<Parser> parsers) {
         this.parsers.clear();
         this.parsers.addAll(parsers);
         return this;
--- a/store/src/main/java/com/nytimes/android/external/store/base/impl/ParsingStoreBuilder.java
+++ b/store/src/main/java/com/nytimes/android/external/store/base/impl/ParsingStoreBuilder.java
@@ -1,6 +1,5 @@
 package com.nytimes.android.external.store.base.impl;
 
-
 import com.nytimes.android.external.cache.Cache;
 import com.nytimes.android.external.store.base.DiskRead;
 import com.nytimes.android.external.store.base.DiskWrite;
@@ -10,11 +9,11 @@ import com.nytimes.android.external.store.base.Persister;
 import com.nytimes.android.external.store.base.Store;
 import com.nytimes.android.external.store.util.NoopPersister;
 
+import org.jetbrains.annotations.Nonnull;
+
 import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.Callable;
-
-import javax.annotation.Nonnull;
 
 import rx.Observable;
 import rx.functions.Func1;
@@ -22,12 +21,12 @@ import rx.functions.Func1;
 /**
  * Builder where there parser is used.
  */
-@SuppressWarnings({"PMD.AvoidFieldNameMatchingMethodName"})
+@Deprecated
 public class ParsingStoreBuilder<Raw, Parsed> {
 
     private final List<Parser> parsers = new ArrayList<>();
-    private Fetcher<Raw> fetcher;
-    private Persister<Raw> persister;
+    private Fetcher<Raw, BarCode> fetcher;
+    private Persister<Raw, BarCode> persister;
     private Cache<BarCode, Observable<Parsed>> memCache;
 
     public ParsingStoreBuilder() {
@@ -35,14 +34,19 @@ public class ParsingStoreBuilder<Raw, Parsed> {
     }
 
     @Nonnull
-    public ParsingStoreBuilder<Raw, Parsed> fetcher(final @Nonnull Fetcher<Raw> fetcher) {
+    public static <Raw, Parsed> ParsingStoreBuilder<Raw, Parsed> builder() {
+        return new ParsingStoreBuilder<>();
+    }
+
+    @Nonnull
+    public ParsingStoreBuilder<Raw, Parsed> fetcher(final @Nonnull Fetcher<Raw, BarCode> fetcher) {
         this.fetcher = fetcher;
         return this;
     }
 
     @Nonnull
     public ParsingStoreBuilder<Raw, Parsed> nonObservableFetcher(final @Nonnull Func1<BarCode, Raw> fetcher) {
-        this.fetcher = new Fetcher<Raw>() {
+        this.fetcher = new Fetcher<Raw, BarCode>() {
             @Nonnull
             @Override
             public Observable<Raw> fetch(final BarCode barCode) {
@@ -59,15 +63,15 @@ public class ParsingStoreBuilder<Raw, Parsed> {
     }
 
     @Nonnull
-    public ParsingStoreBuilder<Raw, Parsed> persister(final @Nonnull Persister<Raw> persister) {
+    public ParsingStoreBuilder<Raw, Parsed> persister(final @Nonnull Persister<Raw, BarCode> persister) {
         this.persister = persister;
         return this;
     }
 
     @Nonnull
-    public ParsingStoreBuilder<Raw, Parsed> persister(final @Nonnull DiskRead<Raw> diskRead,
-                                                      final @Nonnull DiskWrite<Raw> diskWrite) {
-        persister = new Persister<Raw>() {
+    public ParsingStoreBuilder<Raw, Parsed> persister(final @Nonnull DiskRead<Raw, BarCode> diskRead,
+                                                      final @Nonnull DiskWrite<Raw, BarCode> diskWrite) {
+        persister = new Persister<Raw, BarCode>() {
             @Nonnull
             @Override
             public Observable<Raw> read(BarCode barCode) {
@@ -98,15 +102,10 @@ public class ParsingStoreBuilder<Raw, Parsed> {
     }
 
     @Nonnull
-    public ParsingStoreBuilder<Raw, Parsed> parsers(final @Nonnull List<Parser> parsers) {
+    public ParsingStoreBuilder<Raw, Parsed> parsers(final @Nonnull List<Parser<Raw, Parsed>> parsers) {
         this.parsers.clear();
         this.parsers.addAll(parsers);
         return this;
-    }
-
-    @Nonnull
-    public static <Raw, Parsed> ParsingStoreBuilder<Raw, Parsed> builder() {
-        return new ParsingStoreBuilder<>();
     }
 
     @Nonnull
@@ -121,8 +120,8 @@ public class ParsingStoreBuilder<Raw, Parsed> {
             persister = new NoopPersister<>();
         }
 
-        Parser<Raw, Parsed> multiParser = new MultiParser<>(parsers);
-        RealInternalStore<Raw, Parsed> realInternalStore;
+        Parser<Raw, Parsed> multiParser = new MultiParser<Raw, Parsed>(parsers);
+        RealInternalStore<Raw, Parsed, BarCode> realInternalStore;
 
         if (memCache == null) {
             realInternalStore = new RealInternalStore<>(fetcher, persister, multiParser);
@@ -130,6 +129,6 @@ public class ParsingStoreBuilder<Raw, Parsed> {
             realInternalStore = new RealInternalStore<>(fetcher, persister, multiParser, memCache);
         }
 
-        return new RealStore<>(realInternalStore);
+        return new ProxyStore<>(realInternalStore);
     }
 }
--- a/store/src/main/java/com/nytimes/android/external/store/base/impl/ProxyStore.java
+++ b/store/src/main/java/com/nytimes/android/external/store/base/impl/ProxyStore.java
@@ -1,69 +1,65 @@
 package com.nytimes.android.external.store.base.impl;
 
-import javax.annotation.Nonnull;
 
 import com.nytimes.android.external.cache.Cache;
 import com.nytimes.android.external.store.base.Fetcher;
 import com.nytimes.android.external.store.base.InternalStore;
 import com.nytimes.android.external.store.base.Parser;
 import com.nytimes.android.external.store.base.Persister;
-import com.nytimes.android.external.store.base.beta.Store;
+import com.nytimes.android.external.store.base.Store;
 import com.nytimes.android.external.store.util.NoopParserFunc;
 import com.nytimes.android.external.store.util.NoopPersister;
+
+import org.jetbrains.annotations.NotNull;
 
 import rx.Observable;
 import rx.functions.Func1;
 
-public class RealStore<Parsed, Key> implements Store<Parsed, Key> {
+@Deprecated
+public class ProxyStore<Parsed> implements Store<Parsed> {
 
-    private final InternalStore<Parsed, Key> internalStore;
+    private final InternalStore<Parsed, BarCode> internalStore;
 
-    RealStore(InternalStore<Parsed, Key> internalStore) {
+    ProxyStore(InternalStore<Parsed, BarCode> internalStore) {
         this.internalStore = internalStore;
     }
 
-    public RealStore(Fetcher<Parsed, Key> fetcher) {
-        internalStore = new RealInternalStore<>(fetcher, new NoopPersister<Parsed, Key>(),
+    public ProxyStore(Fetcher<Parsed, BarCode> fetcher) {
+        internalStore = new RealInternalStore<>(fetcher, new NoopPersister<Parsed, BarCode>(),
                 new NoopParserFunc<Parsed, Parsed>());
     }
 
-    public RealStore(Fetcher<Parsed, Key> fetcher, Persister<Parsed, Key> persister) {
+    public ProxyStore(Fetcher<Parsed, BarCode> fetcher, Persister<Parsed, BarCode> persister) {
         internalStore = new RealInternalStore<>(fetcher, persister,
                 new NoopParserFunc<Parsed, Parsed>());
     }
 
-    public <Raw> RealStore(Fetcher<Raw, Key> fetcher,
-                           Persister<Raw, Key> persister,
-                           Parser<Raw, Parsed> parser) {
+    public <Raw> ProxyStore(Fetcher<Raw, BarCode> fetcher,
+                            Persister<Raw, BarCode> persister,
+                            Parser<Raw, Parsed> parser) {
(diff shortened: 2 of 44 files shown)


digitalbuddha:
I ran an auto refactor, seems to have fixed some spacing.  Sorry for noise

digitalbuddha:
Leaning towards getting this into a 1.1 release along with #68 that should make the store much more robust.  Gotta keep our users happy :-)

pavlospt:
@digitalbuddha I would also suggest providing a Builder for that `fromTypes()` method, as it is not quite clear from a blackbox perspective.

digitalbuddha:
I'll see if it's possible. Java generics are tough when it comes to builders. What I mean is that I can't do Builder.type(class).type(anotherclass).build. The build method will not have access to the class type due to erasure. Maybe I can do something like .types(BarcodeType(key.class),..)

pavlospt:
@digitalbuddha Couldn't you though do Builder.key(keyClazz).parsed(parsedClazz).raw(rawClazz);` ?

digitalbuddha:
So here's the thing about java....generic support is suboptimal.


Say I have a builder

TypeBuilder(){

...

<T> key(Class<T> keyClass(Class<T> clazz){
      this.keyClass=clazz
}
}

I need to define `keyClass` as 
`Class<T> keyClass;`

But I can only do that if `TypeBuilder` has a generic as well TypeBuilder<T>.  Which then means someone would have to make a builder `TypeBuilder<Key,Raw,Parsed>.keyClass().... defeating the whole purpose of passing in the types.

The only way I saw of getting around this problem is to do 
```java
public static <Key, R, Parsed> BarcodeStoreBuilder<Parsed, R, Key> fromTypes(Class<Key> keyClass,
                                                                                  Class<Parsed> parsedClass,
                                                                                  Class<R> rawClass
     ) {
         return new BarcodeStoreBuilder<>();
     }
```

But yeah Ideally I want to do StoreBuilder.keyType(String.class).rawType(Reader.class), if you know of a way to make that happen I'd love to make a change.  Thanks for reviewing :-)

pavlospt:
@digitalbuddha Why would you need to declare the type though? 

```
public static void main(String[] args) {
        Builder builder = new Builder()
                .key(String.class)
                .parsed(Integer.class)
                .raw(Object.class);

        System.out.println(builder.getKeyClass());
        System.out.println(builder.getParsedClass());
        System.out.println(builder.getRawClass());
    }

    static class Builder {
        private Class keyClass, parsedClass, rawClass;

        public Builder key(Class<? extends String> clazz) {
            this.keyClass = clazz;
            return this;
        }

        public Builder parsed(Class clazz) {
            this.parsedClass = clazz;
            return this;
        }

        public Builder raw(Class clazz) {
            this.rawClass = clazz;
            return this;
        }

        public Class getKeyClass() {
            return keyClass;
        }

        public Class getParsedClass() {
            return parsedClass;
        }

        public Class getRawClass() {
            return rawClass;
        }
    }
```

Wouldn't something like this work ?

digitalbuddha:
The Store that is being built needs to know the type.  Otherwise when you call store.get the key won't be typed to the one you passed in.  I could be missing something though

pavlospt:
@digitalbuddha Ohh now I get it, I missed the point that there was an explicitly declared Type needed!

store/src/main/java/com/nytimes/android/external/store/base/experimental/Store.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public interface Store<T, V> {
+
+    /**
+     * Return an Observable of T for request Barcode
+     * Data will be returned from oldest non expired source
+     * Sources are Memory Cache, Disk Cache, Inflight, Network Response
+     */
+    @NotNull
+    Observable<T> get(@NotNull V barCode);
```
alexio: rename input to `key`, barCode if for old impl

store/src/main/java/com/nytimes/android/external/store/base/experimental/Store.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Return an Observable of T for requested Barcode skipping Memory & Disk Cache
+     */
+    @NotNull
+    Observable<T> fetch(@NotNull V barCode);
```
alexio: rename input to `key`, barCode if for old impl

store/src/main/java/com/nytimes/android/external/store/base/experimental/Store.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Clear the memory cache of all entries
+     */
+    void clearMemory();
+
+    /**
+     * Purge a particular entry from memory cache.
+     */
+    void clearMemory(@NotNull V barCode);
```
alexio: rename input to `key` or `id`, barCode if for old impl

store/src/main/java/com/nytimes/android/external/store/base/experimental/Store.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * a {@link com.nytimes.android.external.store.base.impl.StoreBuilder StoreBuilder}
+ * will return an instance of a store
+ * <p>
+ * A {@link Store  Store} can
+ * {@link Store#get(BarCode) Store.get() } cached data or
+ * force a call to {@link Store#fetch(BarCode) Store.fetch() }
+ * (skipping cache)
+ */
+public interface Store<T, V> {
```
alexio: Why is this in an experimental package and not replacing `com.nytimes.android.external.store.base.Store`? Trying to avoid breaking changes? If so, doesn't adding the 2nd generic to `Fetcher` go against that?

store/src/main/java/com/nytimes/android/external/store/base/FetcherOld.java
```diff
@@ -0,0 +1,24 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import rx.Observable;
+
+/**
+ * Interface for fetching new data for a Store
+ *
+ * @param <Raw> data type before parsing
+ */
+public interface FetcherOld<Raw> extends Fetcher<Raw, BarCode> {
```
alexio: `FetcherOld` doesn't describe the interface properly imo.
Consider renaming to something like `BarcodeFetcher`, etc.

store/src/main/java/com/nytimes/android/external/store/base/Store.java
```diff
@@ -11,40 +11,43 @@
...
- * A {@link com.nytimes.android.external.store.base.Store  Store} can
- * {@link com.nytimes.android.external.store.base.Store#get(BarCode) Store.get() } cached data or
- * force a call to {@link com.nytimes.android.external.store.base.Store#fetch(BarCode) Store.fetch() }
+ * A {@link Store  Store} can
+ * {@link Store#get(BarCode) Store.get() } cached data or
+ * force a call to {@link Store#fetch(BarCode) Store.fetch() }
  * (skipping cache)
  */
-public interface Store<T> {
+public interface Store<T> extends com.nytimes.android.external.store.base.experimental.Store<T, BarCode> {
```
alexio: add import to remove full path.

store/src/main/java/com/nytimes/android/external/store/base/impl/BarcodeStoreBuilder.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import rx.Observable;
+import rx.functions.Func1;
+
+/**
+ * Builder where there parser is used.
+ */
+public class BarcodeStoreBuilder<Raw, Parsed, Key> {
```
alexio: final? Also private constructor.

store/src/main/java/com/nytimes/android/external/store/base/impl/BarcodeStoreBuilder.java
```diff
@@ -0,0 +1,110 @@
...
+            @Override
+            public Observable<Boolean> write(Key barCode, Raw raw) {
+                return diskWrite.write(barCode, raw);
+            }
+        };
+        return this;
+    }
+
+    @NotNull
+    public BarcodeStoreBuilder<Raw, Parsed, Key> parser(final @NotNull Func1<Raw, Parsed> parser) {
```
alexio: this should take in a `Parser<Raw, Parsed>` instead.

alexio: Just noticed that this overloads `parser(final @NotNull Parser<Raw, Parsed> parser)`. What is the need for `parser(final @NotNull Func1<Raw, Parsed> parser)`?

store/src/main/java/com/nytimes/android/external/store/base/impl/BarcodeStoreBuilder.java
```diff
@@ -0,0 +1,110 @@
...
+import java.util.List;
+
+import rx.Observable;
+import rx.functions.Func1;
+
+/**
+ * Builder where there parser is used.
+ */
+public class BarcodeStoreBuilder<Raw, Parsed, Key> {
+    private final List<Parser> parsers = new ArrayList<>();
```
alexio: Should be `List<Parser<Raw, Parsed>>`

store/src/main/java/com/nytimes/android/external/store/base/impl/ProxyStore.java
```diff
@@ -0,0 +1,111 @@
...
+import com.nytimes.android.external.store.base.Store;
+import com.nytimes.android.external.store.util.NoopParserFunc;
+import com.nytimes.android.external.store.util.NoopPersister;
+
+import org.jetbrains.annotations.NotNull;
+
+import rx.Observable;
+import rx.functions.Func1;
+
+public class ProxyStore<Parsed> implements Store<Parsed> {
```
alexio: Class doc would be nice, what's the purpose of this class?

store/src/main/java/com/nytimes/android/external/store/base/impl/StoreBuilder.java
```diff
@@ -13,93 +12,100 @@
...
 
 /**
  * Builder where there parser is used.
  */
-@SuppressWarnings({"PMD.AvoidFieldNameMatchingMethodName"})
-public class StoreBuilder<T> {
-    private Fetcher<T> fetcher;
-    private Persister<T> persister;
-    private Cache<BarCode, Observable<T>> memCache;
+public class StoreBuilder<Raw> {
```
alexio: this should be final as well with a private constructor

alexio: Rename to `BarcodeStore`?

store/src/main/java/com/nytimes/android/external/store/base/experimental/Store.java
digitalbuddha: yes to avoid breaking changes.  So Fetcher will break but at least the StoreBuilder wont as well.  Once we hit version 2, this will be cleaned up.

store/src/main/java/com/nytimes/android/external/store/base/Store.java
alexio: Rename to `BarcodeStore`?

store/src/main/java/com/nytimes/android/external/store/base/impl/BarcodeStoreBuilder.java
digitalbuddha: no need only the Func is needed. thanks!

store/src/main/java/com/nytimes/android/external/store/base/impl/ProxyStore.java
digitalbuddha: backwards compatibility will comment

store/src/main/java/com/nytimes/android/external/store/base/experimental/Store.java
alexio: If you're going to introduce breaking changes anyway, just do it all in one go. Reduce tech debt.

PaulWoitaschek:
Why do you want to keep the whole BarCode class anyway?

store/src/main/java/com/nytimes/android/external/store/base/impl/BarcodeStoreBuilder.java
alexio: Remove the `parser(final @NotNull Func1<Raw, Parsed> parser)` and keep `parser(final @NotNull Parser<Raw, Parsed> parser)`? For type safety :).

digitalbuddha:
@PaulWoitaschek because its being used internally in both our apps.  The whole breaking changes thing....Was hoping to minimize.

alexio:
I also think that the BarCode class isn't necessary. Breaking changes are being made anyway, just do it in one sweep?

PaulWoitaschek:
You could just copy the current `BarCode` class to your project and use it as the key type and change the imports. Might be even less overhead and reduce general complexity of the library.

store/src/main/java/com/nytimes/android/external/store/base/experimental/Store.java
digitalbuddha: disclaimer was added to readme up top. good suggestion

alexio:
agree with @PaulWoitaschek

store/src/main/java/com/nytimes/android/external/store/base/Store.java
digitalbuddha: renaming would add to breaking changes, I'll just throw a deprecated tag up here

store/src/main/java/com/nytimes/android/external/store/base/impl/BarcodeStoreBuilder.java
digitalbuddha: sure

digitalbuddha:
Sure but this library is being used externally as well. I'll leave Barcode in for version 1 and remove in version 2. Let's shoot for a v2 release in early April.  Its a few dozen methods so it shouldn't have any real affect on anyone if it is left in.

PaulWoitaschek:
Give all that stuff a `@Deprecated` annotation if it gets removed soon?

digitalbuddha:
yup doing now

store/src/main/java/com/nytimes/android/external/store/base/impl/BarcodeStoreBuilder.java
digitalbuddha: actually parser needs to be this way because you may be passing in 2 parsers, one of which is <String,Integer> and the other that is <Integer,Long> so the Parser list doesn't need to concern itself with what the types are of the individual parsers. We have a cast exception at runtime if it doesn't all line up :-)

digitalbuddha:
sorry to do this to y'all just rebased and fixed all the comments, mind reviewing again @brianPlummer @alexio

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -351,20 +351,21 @@ private long getCacheTTL() {
      * @return memory persister size
      */
     private long getCacheSize() {
+//        return memCache.size();
```
brianPlummer: intentional???

digitalbuddha: Hmm weird no idea. Need to look at old commits.

store/src/main/java/com/nytimes/android/external/store/base/impl/ParsingStoreBuilder.java
```diff
@@ -22,27 +21,32 @@
 /**
  * Builder where there parser is used.
  */
-@SuppressWarnings({"PMD.AvoidFieldNameMatchingMethodName"})
+@Deprecated
 public class ParsingStoreBuilder<Raw, Parsed> {
 
     private final List<Parser> parsers = new ArrayList<>();
```
alexio: changed to `List<Parser<Raw, Parsed>>`

store/src/main/java/com/nytimes/android/external/store/base/impl/RealStoreBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+
+import javax.annotation.Nonnull;
+
+import rx.Observable;
+
+/**
+ * Builder where there parser is used.
+ */
+public class RealStoreBuilder<Raw, Parsed, Key> {
+    private final List<Parser> parsers = new ArrayList<>();
```
alexio: change to `List<Parser<Raw,Parsed>>`

store/src/main/java/com/nytimes/android/external/store/base/impl/RealStoreBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @Nonnull
+    public RealStoreBuilder<Raw, Parsed, Key> parser(final @Nonnull Parser<Raw, Parsed> parser) {
+        this.parsers.clear();
+        this.parsers.add(parser);
+        return this;
+    }
+
+    @Nonnull
+    public RealStoreBuilder<Raw, Parsed, Key> parsers(final @Nonnull List<Parser> parsers) {
```
alexio: change to `List<Parser<Raw,Parsed>>`

store/src/main/java/com/nytimes/android/external/store/base/impl/StoreBuilder.java
```diff
@@ -11,96 +10,107 @@
...
 
 /**
  * Builder where there parser is used.
  */
-@SuppressWarnings({"PMD.AvoidFieldNameMatchingMethodName"})
-public class StoreBuilder<T> {
-    private Fetcher<T> fetcher;
-    private Persister<T> persister;
-    private Cache<BarCode, Observable<T>> memCache;
+public final class StoreBuilder<Raw> {
```
alexio: add `@Deprecated`

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -239,12 +231,13 @@ public void call(Parsed parsed) {
...
 
     @Nonnull
-    Observable<Parsed> response(@Nonnull final BarCode barCode) {
+    Observable<Parsed> response(@Nonnull final Key barCode) {
         return fetcher()
                 .fetch(barCode)
                 .flatMap(new Func1<Raw, Observable<Parsed>>() {
                     @Override
                     public Observable<Parsed> call(Raw raw) {
+                        //Log.i(TAG,"writing and then reading from Persister");
```
alexio: remove or uncomment log?

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -261,16 +254,17 @@ public void call(Parsed data) {
...
-                    public void call(Throwable throwable) {
+                    public void call() {
                         inFlightRequests.invalidate(barCode);
                     }
                 })
                 .cache();
     }
 
     void notifySubscribers(Parsed data) {
+        //Log.d(TAG,"notify stream subscribers of fresh data");
```
alexio: remove or uncomment?

PaulWoitaschek:
What's the reason for passing the classes in? It creates quite some overhead while creating the stores. Can't you just give them different names? I now need to pass in a lot of class objects.

```java
public static <Parsed> RealStoreBuilder<Parsed, Parsed, BarCode> barcode() {
    return new RealStoreBuilder<>();
  }

  public static <Key, Parsed> RealStoreBuilder<Parsed, Parsed, Key> key() {
    return new RealStoreBuilder<>();
  }

  @Beta
  public static <Key, Raw, Parsed> RealStoreBuilder<Raw, Parsed, Key> parsedWithKey() {
    return new RealStoreBuilder<>();
  }
```

digitalbuddha:
Good idea will change

PR #90, opened 2017-01-30
fixes race condition in memory cache

digitalbuddha:
closes  89 thanks @PaulWoitaschek for finding and submitting a clear concise test case.


digitalbuddha:
OK finally found the root cause. Internally we use a lazy Persister.  The current impl of the `NoOpPersister` was not lazy. Wrapped all calls to `disk` to be lazy. For safeguard wrapped that calls to `memcache` as well.

PR #94, opened 2017-01-31
use javax annotations instead of intellij

brianPlummer:


cache/src/main/java/com/nytimes/android/external/cache/AbstractFuture.java
```diff
@@ -16,6 +13,9 @@
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import javax.annotation.Nonnull;
```
digitalbuddha: every single one has a different capitalization smh

brianPlummer: sounds legit....why have it one way when everyone can do what they want ;)

PR #97, opened 2017-01-31
clear needs to clear noop disk and inflight

### LINKED ISSUE #96: Clear Memory Fails
After clearing the memory a following call to `get` should make a new request. But it doesn't

See #95

digitalbuddha:
closes #96  
@PaulWoitaschek you find em I fix em? :-)

### DIFF
--- a/store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
+++ b/store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
@@ -265,9 +265,9 @@ final class RealInternalStore<Raw, Parsed> implements InternalStore<Parsed> {
 
     @Override
     public void clearMemory() {
-        memCache.invalidateAll();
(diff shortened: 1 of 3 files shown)


store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -265,6 +266,14 @@ void updateMemory(@Nonnull final BarCode barCode, final Parsed data) {
     @Override
     public void clearMemory() {
         memCache.invalidateAll();
+        inFlightRequests.invalidateAll();
```
pavlospt: Shouldn't you clear the inflight requests first in case there is one that successes and writes in memory after you have cleared it and before you invalidate the inFlight ones?

digitalbuddha: Yup you're right

PR #99, opened 2017-01-31
Remove dead code and add/remove empty lines were needed

pavlospt:
Just some "style" fixes I would say. Still not searched through the entire project, will do over the next days and fix some more if I find.

### DIFF
--- a/store/src/main/java/com/nytimes/android/external/store/base/impl/BarCode.java
+++ b/store/src/main/java/com/nytimes/android/external/store/base/impl/BarCode.java
@@ -43,14 +43,17 @@ public final class BarCode implements Serializable {
         if (this == object) {
             return true;
         }
+
         if (!(object instanceof BarCode)) {
             return false;
         }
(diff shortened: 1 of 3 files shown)


PaulWoitaschek:
Btw, what about lambdas in general?

pavlospt:
@PaulWoitaschek Can you explain a bit more ? :)

PaulWoitaschek:
I would love to see lambdas in the source all over the place. Not directly related to your pull request but not that I would open an issue for.

digitalbuddha:
@PaulWoitaschek I wanted to keep adoption as high as possible. There are apps that do not use Retrolambda yet.  Before release I actually had to remove all the lambdas :-(

PaulWoitaschek:
Is it actually an issue for libraries? Don't they get compiled to bytecode targeting JDK6?

digitalbuddha:
I can take a look good thought....

PR #103, opened 2017-02-02
GetRefreshing - Repeat when clear

### LINKED ISSUE #68: Repeat When Clear functionality

digitalbuddha:
Closes #68 

@PaulWoitaschek lmk if any feedback.  Marking as experimental for now

### DIFF
--- a/store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
+++ b/store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
@@ -43,11 +43,11 @@ final class RealInternalStore<Raw, Parsed> implements InternalStore<Parsed> {
     Cache<BarCode, Observable<Parsed>> inFlightRequests;
     Cache<BarCode, Observable<Parsed>> memCache;
 
+    private final PublishSubject<BarCode> refreshSubject = PublishSubject.create();
     private Fetcher<Raw> fetcher;
     private Persister<Raw> persister;
     private Func1<Raw, Parsed> parser;
     private BehaviorSubject<Parsed> subject;
-    private PublishSubject<BarCode> refreshSubject = PublishSubject.create();
 
     RealInternalStore(Fetcher<Raw> fetcher,
                       Persister<Raw> persister,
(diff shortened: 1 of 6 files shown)


PaulWoitaschek:
I think it should respond to `clearAll` events too.

An easy way would be to have a second PublishSubject and merge its emissions with the `Observable<Barcode> filter>`.

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -264,6 +308,7 @@ public void clearMemory() {
         inFlightRequests.invalidateAll();
         clearDiskIfNoOp();
         memCache.invalidateAll();
+        refreshSubject.onNext(ALL_BARCODE);
```
digitalbuddha: got away with using a single subject but filtering on a particular Barcode that signals to clear all

PaulWoitaschek: Great! Make it private?

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -90,6 +96,44 @@ private void init(Fetcher<Raw> fetcher,
...
+    public Observable<Parsed> getRefreshing(@Nonnull final BarCode barCode) {
+        return get(barCode).compose(repeatWhenCacheEvicted(barCode));
+    }
+
+    @Nonnull
+    private Observable.Transformer<Parsed, Parsed> repeatWhenCacheEvicted(@Nonnull final BarCode key) {
+        Observable<BarCode> filter = refreshSubject.filter(new Func1<BarCode, Boolean>() {
+            @Override
+            public Boolean call(BarCode barCode) {
+                return key.equals(barCode) || barCode.equals(ALL_BARCODE);
```
pavlospt: @digitalbuddha For consistency you could say 

```java
barcode.equals(key) || barCode.equals(ALL_BARCODE)
```

digitalbuddha:
@PaulWoitaschek this is the first feature that we are not using internally. Let me know if anything is broken :-) thanks for suggesting it, I'll have a snapshot out tomorrow.

Issue #69, opened 2017-01-18
Add record for File System


digitalbuddha:
A Record represents metadata about a stored object, kind of like a java 'File'
or 'stat' in a unix filesystem. The idea here is to allow you to get info like
last modified time without actually reading the file's contents.

PR #111, opened 2017-02-06
networkBeforeStale and refreshOnStale

digitalbuddha:
This PR introduces 2 new StoreBuilder configurations:
  
  **refreshOnStale()**  -Store will backfill the disk cache anytime a record is stale, User will still get the stale record returned to them

  **networkBeforeStale**  - Store will try to get network source when disk data is stale if network source throws error or is empty, stale disk data will be returned

This PR also implements a new FileSystem middleware - **RecordPersister**, users can implement `RecordProvider`to make their own `Persister` record aware

Tests are a good place to start when reviewing
StoreNetworkBeforeStaleTest.java && StoreRefreshWhenStaleTest.java


digitalbuddha:
Closes #69

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -176,17 +151,44 @@ public Boolean call(Key barCode) {
...
+            @Override
+            public void call(Parsed parsed) {
+                //do Nothing we are just backfilling cache
+            }
+        }, new Action1<Throwable>() {
+            @Override
+            public void call(Throwable throwable) {
+                //do nothing as we are just backfilling cache
+            }
+        });
```
cybo42: Maybe create a "noop" `Action1` to use here for readability instead of the empty / comment implementations. Could also be reused if needed elsewhere

Ex.
```
        Action1 noop = new Action1(){
            @Override
            public void call(Object o) {
            // do nothing
            }
        };
...
        fetch(barCode).subscribe(noop, noop);
```

PR #117, opened 2017-02-08
Feature/clear all cache

### LINKED ISSUE #102: Add ClearDisk interface
in conjunction with #68 we should have a way to clear memory and disk cache.  Current idea is add:

Store:
`    void clear(BarCode barCode);
`
Builder:
`   builder.diskClearer(barcode->//deletes entry from user's persister);`

digitalbuddha:
close #102 

This PR introduces a `Clearable` interface & 2 public methods in store: `clear(key)/clear()` which works as:

If a `Persister` instance implements `Clearable` then the disk cache will be cleared otherwise memory only will be cleared. The functionality we are trying to achieve is clearing all caches allowing for getRefreshing to truly sync back to the server.

### DIFF
--- a/store/src/main/java/com/nytimes/android/external/store/base/Clearable.java
+++ b/store/src/main/java/com/nytimes/android/external/store/base/Clearable.java
@@ -1,9 +1,10 @@
 package com.nytimes.android.external.store.base;
 
-/**
- * Created by 206847 on 2/7/17.
- */
 
+/**
+ * Persisters should implement Clearable if they want store.clear(key) to also clear the persister
+ * @param <T> Type of key/request param in store
+ */
 public interface Clearable<T> {
     void clear(T key);
 }
--- a/store/src/main/java/com/nytimes/android/external/store/base/Clearable.java
+++ b/store/src/main/java/com/nytimes/android/external/store/base/Clearable.java
@@ -1,0 +1,9 @@
+package com.nytimes.android.external.store.base;
+
+/**
+ * Created by 206847 on 2/7/17.
+ */
+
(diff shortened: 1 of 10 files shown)


store/src/main/java/com/nytimes/android/external/store/base/Clearable.java
```diff
@@ -0,0 +1,9 @@
+package com.nytimes.android.external.store.base;
+
+/**
+ * Created by 206847 on 2/7/17.
+ */
+
+public interface Clearable<T> {
```
alexio: interface/method comments :)

store/src/main/java/com/nytimes/android/external/store/base/beta/Store.java
```diff
@@ -63,12 +64,21 @@
...
+    @Deprecated
     void clearMemory();
 
     /**
      * Purge a particular entry from memory cache.
      */
+    @Deprecated
     void clearMemory(@Nonnull V key);
 
+    void clear();
```
alexio: Add a comment so that users will know how clear works. Clears memory, but also clears disk when persister implements clearable.

store/src/main/java/com/nytimes/android/external/store/base/beta/Store.java
```diff
@@ -63,12 +64,21 @@
...
+
+    /**
+     * Purge a particular entry from memory and disk cache.
+     */
+    void clear(@Nonnull V key);
```
alexio: also mention that disk cache is only cleared if persister implements `Clearable`

PR #121, opened 2017-02-08
Fixed a race condition in FS

PaulWoitaschek:
…  that would appear when the read was called but not immediately subscribed to. 

Defer the whole file check until subscription.


digitalbuddha:
Yup! thank you good sir.

PR #124, opened 2017-02-09
Create single instance of empty BarCode + naming cleanup

pavlospt:
Moved empty Strings to private static final variables

### DIFF
--- a/store/src/main/java/com/nytimes/android/external/store/base/impl/BarCode.java
+++ b/store/src/main/java/com/nytimes/android/external/store/base/impl/BarCode.java
@@ -12,6 +12,10 @@ import javax.annotation.Nonnull;
  * and {@link com.nytimes.android.external.store.base.Persister  Persister}
  **/
 public final class BarCode implements Serializable {
+
+    private static final String EMPTY_TYPE = "";
+    private static final String EMPTY_KEY = "";
+
     @Nonnull
     private final String key;
     @Nonnull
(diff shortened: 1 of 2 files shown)


store/src/main/java/com/nytimes/android/external/store/base/impl/BarCode.java
```diff
@@ -24,7 +28,7 @@ public BarCode(@Nonnull String type, @Nonnull String key) {
 
     @Nonnull
     public static BarCode empty() {
-        return new BarCode("", "");
```
PaulWoitaschek: BarCode is immutable, better create a static instancd of the empty barcode and return that.

pavlospt:
@digitalbuddha Yeah I got the project open now and just fixing things, expect some more for today 😄

digitalbuddha:
I plan to cut 2.0 early next week. Lmk if any name/api change are needed
before we go live.

store/src/main/java/com/nytimes/android/external/store/base/impl/BarCode.java
pavlospt: @PaulWoitaschek  Thanks! Done!

pavlospt:
@digitalbuddha Yeah will be on time till then! No worries!

digitalbuddha:
I don't think we need the other static strings. Especially if they are the same.

pavlospt:
@digitalbuddha I did in case an empty type and empty key have different definitions in the future!

PaulWoitaschek:
Thats just an implementation detail and does not make the code more readable or maintainable. In that case the shorter version is better.

Especially now that you eliminated the unnecessary object creation by making a single instance of the empty barcode.

pavlospt:
@PaulWoitaschek Ok great then!

PR #128, opened 2017-02-10
Test repeat when evicted

PaulWoitaschek:
* Moved repeat logic to a separate transformer
* Test that transformer


PaulWoitaschek:
Hm; what can I do to make git not show the merged commits?

pavlospt:
@PaulWoitaschek How are you merging?

digitalbuddha:
Instead of merging, you should rebase develop and then git push -f

pavlospt:
Yeah that's true! Actually do a `git push` of your commit then `git rebase upstream/develop` and then `git push -f` your own develop, so that it will update the PR.

PaulWoitaschek:
That worked

store/src/main/java/com/nytimes/android/external/store/base/impl/RepeatWhenEmits.java
```diff
@@ -0,0 +1,41 @@
...
+        this.source = source;
+    }
+
+    @Nonnull
+    static <T> RepeatWhenEmits<T> from(@Nonnull Observable source) {
+        return new RepeatWhenEmits<>(checkNotNull(source));
+    }
+
+    @Override
+    public Observable<T> call(Observable<T> upstream) {
```
digitalbuddha: Still can barely understand this part. Thanks for writing test to make it clearer and for fixing my error.

PaulWoitaschek: It was not your error, it was mine while renaming the signatures.

For understanding check [this post](http://blog.danlew.net/2016/01/25/rxjavas-repeatwhen-and-retrywhen-explained/) from Dan Lew.

PR #131, opened 2017-02-18
Manually instantiate UTF-8 as a charset instead of using StandardCharsets

stoyicker:
As `StandardCharsets` is only available in API 19 and above.


digitalbuddha (approved):

stoyicker:
I think I'm going to go to bed and tomorrow I'll do this from local instead.

digitalbuddha:
Np! Thanks for finding. Mind making an issue for tracking?

PR #134, opened 2017-02-19
Do not depend on lint task if it does not exist.

stoyicker:
The `java` plugin does not have a `lint` task, so trying to build the project locally fails for me because it is not found. I don't know how Travis has managed to make it work so far tho


digitalbuddha:
Thanks for the contribution!

PR #136, opened 2017-02-21
add KeyParseFunc for times when parser needs input val

digitalbuddha:
This feature adds a new builder method `.fetcher(KeyParseFunc<Key,Raw,Parsed>)` which will allow us to have parsers with access to the input param. This work is needed for `SectionFrontStore`

### DIFF
--- a/store/src/main/java/com/nytimes/android/external/store/util/NoKeyParseFunc.java
+++ b/store/src/main/java/com/nytimes/android/external/store/util/NoKeyParseFunc.java
@@ -1,0 +1,16 @@
+package com.nytimes.android.external.store.util;
+
+import com.nytimes.android.external.store.base.Parser;
+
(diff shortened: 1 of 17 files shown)


store/src/main/java/com/nytimes/android/external/store/util/NoKeyParseFunc.java
```diff
@@ -0,0 +1,16 @@
+package com.nytimes.android.external.store.util;
+
+import com.nytimes.android.external.store.base.Parser;
+
+public class NoKeyParseFunc<Key,Raw,Parsed> implements KeyParseFunc<Key, Raw,Parsed> {
+    private Parser<Raw, Parsed> parser;
```
brianPlummer: can be final

brianPlummer:
lgtm

PR #138, opened 2017-02-23
Introduce RecordPersisterFactory

charbgr:
As there is a SourcePersisterFactory, I thought it would be great if there is also one for the RecordPersister.

Let me know what do you think.


digitalbuddha:
Thanks for contribution! I just merged another `RecordPersister` impl FileSystemRecordPersister mind adding a factory for that as well?

pavlospt:
@digitalbuddha Since this is something that we also need, I am making a PR for it. Do you have any ETA when these could be released on Maven?

digitalbuddha:
In that case let me merge this PR first and then you can make yours.  `2.0.1-SNAPSHOT` will be up within 15 min of me merging the pr.  If you need a  2.0.1 stable release I can do one monday.

PR #144, opened 2017-03-01
modify clearMemory() to call regular clear() since they are equivalant…

brianPlummer:
..,  bullet-proof BreadthFirstFileTreeIterator


filesystem/src/main/java/com/nytimes/android/external/fs/filesystem/BreadthFirstFileTreeIterator.java
```diff
@@ -51,7 +52,9 @@
      * @param root The root directory
      */
     BreadthFirstFileTreeIterator(@Nonnull File root) {
```
digitalbuddha, line 55: Since we got this wrong the first time let's wrap the logic in tests to make sure it won't blow up again.

digitalbuddha:
We should mention in next release that clearMemory is going Dunzo soon

filesystem/src/main/java/com/nytimes/android/external/fs/filesystem/BreadthFirstFileTreeIterator.java
```diff
@@ -51,7 +52,9 @@
...
-        this.currentList = root.listFiles();
+        if (root.listFiles() != null) {
```
pavlospt: Better keep this in a single variable, in case it is a quite huge dir and it is expensive to list the files each time?

```java
File[] listedFiles = root.listFiles();
if(listedFiles != null){
    currentList = Arrays.copy(listedFiles, listedFiles.length);
}
```

brianPlummer: good catch! will do!

filesystem/src/main/java/com/nytimes/android/external/fs/filesystem/BreadthFirstFileTreeIterator.java
brianPlummer: done!

digitalbuddha:
Wait til morning for the international crowd in case someone sees something else

digitalbuddha:
Lgtm after checkstyle fix

PR #146, opened 2017-03-06
Remove dagger dependency and clean unused dependencies from modules

aminelaadhari:
Fix: Some modules depend on unused dependencies or some dependencies should be testComplie
Remove dagger dependency form modules

### DIFF
--- a/buildsystem/dependencies.gradle
+++ b/buildsystem/dependencies.gradle
@@ -49,6 +49,7 @@ ext.versions = [
         autoService          : '1.0-rc2',
         autoCommon           : '0.6',
         javaWriter           : '2.5.1',
+        javax                : '1',
 
         // Debugging & Inspecting.
         slf4j                : '1.7.19',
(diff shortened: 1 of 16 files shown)


digitalbuddha:
I think we need to make a few changes internally before merging this as we rely on those @Inject annotations on constructors.  Thank you for cleanup! will check and merge later today :-)

digitalbuddha:
Could you leave the inject annotations in. I think those are Jsr and not dagger.

aminelaadhari:
Yes true, I will do that.

digitalbuddha:
Also Pls run check task locally. Looks like some Pmd failures. Otherwise lgtm I'll merge when fixed.

aminelaadhari:
Done. I removed the checkNotNull in Util.createParentDirs because the parameter is annotated with Nonnull. I think the check is not useful anymore.
The pmd was failing because checkNotNull throws a NullPointerException.

alexio (approved):

Issue #148, opened 2017-03-07
Move Store in sample out of activity. 


digitalbuddha:
Current sample is weird because the store gets recreated on rotate. We should move store creation to the application class.

Issue #147, opened 2017-03-07
Why fetch is called every time when getting datas?


tsaisean:
I'm using `get` method, but `fetch` still get called every time.
I don't know the cache is working or not? 🤔 

```
@Override
protected void onResume() {
    super.onResume();
    Log.d(TAG, "onResume");

    loadPosts();
}
```

```
public void loadPosts() {
    Log.d(TAG, "[loadPosts] start");

    BarCode awwRequest = new BarCode(RedditData.class.getSimpleName(), "aww");

    Log.d(TAG, "[loadPosts] Call provideRedditStore get with key \"aww\"");

    provideRedditStore()
            .get(awwRequest)
            .subscribeOn(Schedulers.io())
            .observeOn(AndroidSchedulers.mainThread())
            .subscribe(new Subscriber<Integer>() {
                @Override
                public void onCompleted() {
                    Log.d(TAG, "[onCompleted]");
                }

                @Override
                public void onError(Throwable e) {

                }

                @Override
                public void onNext(Integer posts) {
                    Log.d(TAG, "[onNext] get " + posts);
                }
            });
}
```

```
private int i = 1;
private Store<Integer, BarCode> provideRedditStore() {
    return StoreBuilder.<Integer>barcode()
            .fetcher(new Fetcher<Integer, BarCode>() {
                @Nonnull
                @Override
                public Observable<Integer> fetch(@Nonnull BarCode barCode) {
                    try {
                        Thread.sleep(5000);
                    } catch (InterruptedException e) {
                        e.printStackTrace();
                    }
                    int current_i = StoreActivity.this.i++;
                    Log.d(TAG, "[fetch] return " + current_i);
                    return Observable.just(current_i);
                }
            })
            .open();
}
```

pavlospt:
@tsaisean Can you also please post the `get` part of your code?

tsaisean:
@pavlospt Updated.

pavlospt:
@tsaisean And what version are you using?

tsaisean:
@pavlospt I'm using the sample inside the **Store** project. "Store/app/src/"

pavlospt:
@tsaisean Did you delete the persister or something like that? The first time it is supposed to request from Network due to a cache miss, but the second time it will hit your cache.

pavlospt:
[...]

PR #149, opened 2017-03-07
Fix sample app to init Store on Application creation

pavlospt:
Fixing #147 & #148

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -8,9 +8,10 @@
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:supportsRtl="true"
+        android:name=".SampleApp"
         android:theme="@style/AppTheme">
         <activity
-            android:name=".activity.PersistingStoreActivity">
+            android:name=".activity.StoreActivity">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN"/>
 
--- a/app/src/main/java/com/nytimes/android/sample/SampleApp.java
+++ b/app/src/main/java/com/nytimes/android/sample/SampleApp.java
@@ -1,0 +1,100 @@
+package com.nytimes.android.sample;
+
+import android.app.Application;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.nytimes.android.external.cache.CacheBuilder;
+import com.nytimes.android.external.fs.SourcePersisterFactory;
+import com.nytimes.android.external.store.base.Persister;
+import com.nytimes.android.external.store.base.impl.BarCode;
+import com.nytimes.android.external.store.base.impl.Store;
+import com.nytimes.android.external.store.base.impl.StoreBuilder;
+import com.nytimes.android.external.store.middleware.GsonParserFactory;
+import com.nytimes.android.sample.data.model.GsonAdaptersModel;
+import com.nytimes.android.sample.data.model.RedditData;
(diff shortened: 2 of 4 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -8,9 +8,10 @@
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:supportsRtl="true"
+        android:name=".SampleApp"
         android:theme="@style/AppTheme">
         <activity
-            android:name=".activity.PersistingStoreActivity">
```
digitalbuddha, line 13: Don't we need both activities in manifest?

digitalbuddha: Oh sorry that was just label duh

digitalbuddha (approved):

app/src/main/AndroidManifest.xml
pavlospt: Yeah it's just a matter of what a user wants to test :)

digitalbuddha (approved):

alexio (approved):

app/src/main/java/com/nytimes/android/sample/SampleApp.java
```diff
@@ -0,0 +1,100 @@
...
+    public Store<RedditData, BarCode> getPersistedStore() {
+        return this.persistedStore;
+    }
+
+    private Store<RedditData, BarCode> provideRedditStore() {
+        return StoreBuilder.<RedditData>barcode()
+                .fetcher(barCode -> provideRetrofit().fetchSubreddit(barCode.getKey(), "10"))
+                .memory(CacheBuilder.newBuilder()
+                        .maximumSize(1)
+                        .expireAfterWrite(10, TimeUnit.SECONDS)
```
tsaisean, line 63: @pavlospt 
The setting has no effect. The cache won't get cleared after 10 secs.
Still can get data from memory after 10 secs.

digitalbuddha: I know why, I'll fix it in am.

pavlospt: @tsaisean it works fine for me. It didnt go through the fetcher during the 10 seconds period.

tsaisean: @pavlospt What about after 10secs?

pavlospt: @tsaisean just saw that one. Since @digitalbuddha said he knows what's wrong, I think we should wait till he replies and someone can get on with a fix!

digitalbuddha: I need to double check but there's a few things working against this example. First there is a 1min denouncer in the network layer. Second there is a no op persister that is caching stuff in memory. Will have a fix today. Nothing should be affected for real cache policies (over 1 min)

pavlospt: @digitalbuddha maybe we'll need to adjust the debouncer to the cache TTL ?

digitalbuddha: Yes also the noop persister is currently backed by a hashmap, that also needs to be a guava cache with timeout same as memory cache and denouncer. So really denouncer and noop persister should have a max of the mem cache policy.

PR #151, opened 2017-03-08
Fix NoopPersister wrong caching duration

### LINKED ISSUE #150: MO no op MO problems
#149 raises another interesting issue. If the mem cache is less than a minute the in flight denouncer also needs to be adjusted. Secondly the no op persister needs to be migrated to use a guava cache which has same policy as mem cache.

pavlospt:
Attempt to fix #150. Seems to be working from my testing. Let me know what you think and what could have been done better!

### DIFF
--- a/app/src/main/java/com/nytimes/android/sample/SampleApp.java
+++ b/app/src/main/java/com/nytimes/android/sample/SampleApp.java
@@ -1,11 +1,13 @@
 package com.nytimes.android.sample;
 
 import android.app.Application;
+import android.util.Log;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.nytimes.android.external.cache.CacheBuilder;
 import com.nytimes.android.external.fs.SourcePersisterFactory;
+import com.nytimes.android.external.store.base.Fetcher;
 import com.nytimes.android.external.store.base.Persister;
 import com.nytimes.android.external.store.base.impl.BarCode;
 import com.nytimes.android.external.store.base.impl.Store;
@@ -17,6 +19,8 @@ import com.nytimes.android.sample.data.remote.Api;
 
 import java.io.IOException;
 import java.util.concurrent.TimeUnit;
+
+import javax.annotation.Nonnull;
 
 import okio.BufferedSource;
 import retrofit2.GsonConverterFactory;
@@ -57,11 +61,16 @@ public class SampleApp extends Application {
 
     private Store<RedditData, BarCode> provideRedditStore() {
         return StoreBuilder.<RedditData>barcode()
-                .fetcher(barCode -> provideRetrofit().fetchSubreddit(barCode.getKey(), "10"))
-                .memory(CacheBuilder.newBuilder()
-                        .maximumSize(1)
-                        .expireAfterWrite(10, TimeUnit.SECONDS)
-                        .build())
+                .fetcher(new Fetcher<RedditData, BarCode>() {
+                    @Nonnull
+                    @Override
+                    public Observable<RedditData> fetch(@Nonnull BarCode barCode) {
+                        Log.d("fetcher", "Going for fetcher");
+                        return provideRetrofit().fetchSubreddit(barCode.getKey(), "10");
+                    }
+                })
+//                .fetcher(barCode -> provideRetrofit().fetchSubreddit(barCode.getKey(), "10"))
+                .memory(10, TimeUnit.SECONDS)
                 .open();
     }
 
--- a/app/src/main/java/com/nytimes/android/sample/SampleApp.java
+++ b/app/src/main/java/com/nytimes/android/sample/SampleApp.java
@@ -5,11 +5,11 @@ import android.util.Log;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
-import com.nytimes.android.external.cache.CacheBuilder;
 import com.nytimes.android.external.fs.SourcePersisterFactory;
 import com.nytimes.android.external.store.base.Fetcher;
 import com.nytimes.android.external.store.base.Persister;
 import com.nytimes.android.external.store.base.impl.BarCode;
+import com.nytimes.android.external.store.base.impl.MemoryPolicy;
 import com.nytimes.android.external.store.base.impl.Store;
 import com.nytimes.android.external.store.base.impl.StoreBuilder;
 import com.nytimes.android.external.store.middleware.GsonParserFactory;
@@ -70,7 +70,14 @@ public class SampleApp extends Application {
                     }
                 })
 //                .fetcher(barCode -> provideRetrofit().fetchSubreddit(barCode.getKey(), "10"))
-                .memory(10, TimeUnit.SECONDS)
+                .memoryPolicy(
+                    MemoryPolicy
+                        .MemoryPolicyBuilder
+                        .newBuilder()
+                        .setExpireAfter(10)
+                        .setExpireAfterTimeUnit(TimeUnit.SECONDS)
+                        .build()
+                )
                 .open();
     }
 
--- a/app/src/main/java/com/nytimes/android/sample/SampleApp.java
+++ b/app/src/main/java/com/nytimes/android/sample/SampleApp.java
@@ -61,15 +61,7 @@ public class SampleApp extends Application {
 
     private Store<RedditData, BarCode> provideRedditStore() {
         return StoreBuilder.<RedditData>barcode()
-                .fetcher(new Fetcher<RedditData, BarCode>() {
-                    @Nonnull
-                    @Override
-                    public Observable<RedditData> fetch(@Nonnull BarCode barCode) {
-                        Log.d("fetcher", "Going for fetcher");
-                        return provideRetrofit().fetchSubreddit(barCode.getKey(), "10");
-                    }
-                })
-//                .fetcher(barCode -> provideRetrofit().fetchSubreddit(barCode.getKey(), "10"))
+                .fetcher(barCode -> provideRetrofit().fetchSubreddit(barCode.getKey(), "10"))
                 .memoryPolicy(
                     MemoryPolicy
                         .builder()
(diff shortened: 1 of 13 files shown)


digitalbuddha (commented):
wow thank you so much for this! Ok so theres 1 more whole you can configure the cache to have `                .maximumSize()` Ideally the no op persister would respect this limit as well.  Since we have 3 fields to pass through lets make a `MemoryConfig` java object that will have the 3 fields "size, duration, unit" or some similar name.  Thank you again overall looks solid.

pavlospt:
@digitalbuddha yeah that sounds great. Would definitely do that tomorrow and push again!

pavlospt:
Will also fix the Travis issues on the same commit!

app/src/main/java/com/nytimes/android/sample/SampleApp.java
```diff
@@ -57,11 +61,16 @@ private void initPersister() {
...
+                .fetcher(new Fetcher<RedditData, BarCode>() {
+                    @Nonnull
+                    @Override
+                    public Observable<RedditData> fetch(@Nonnull BarCode barCode) {
+                        Log.d("fetcher", "Going for fetcher");
+                        return provideRetrofit().fetchSubreddit(barCode.getKey(), "10");
+                    }
+                })
+//                .fetcher(barCode -> provideRetrofit().fetchSubreddit(barCode.getKey(), "10"))
+                .memory(10, TimeUnit.SECONDS)
```
digitalbuddha: lets change method name to `memoryPolicy`

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -53,41 +53,60 @@
...
+        this.fetcher = fetcher;
+        this.persister = persister;
+        this.parser = parser;
+        this.memCache = CacheBuilder
+                .newBuilder()
+                .expireAfterWrite(expireAfter, expireAfterTimeUnit)
+                .build();
+        this.stalePolicy = stalePolicy;
+        this.inFlightRequests = CacheBuilder.newBuilder()
+                .expireAfterWrite(expireAfter, expireAfterTimeUnit)
```
digitalbuddha: in flight request should not necessarily be same as memory, its more that it should be the max(1 minute, memory limit)

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -53,41 +53,60 @@
...
         this.parser = parser;
-        this.memCache = memCache;
+        this.memCache = CacheBuilder
+                .newBuilder()
+                .expireAfterWrite(getCacheTTL(), getCacheTTLTimeUnit())
+                .build();
         this.stalePolicy = stalePolicy;
-        inFlightRequests = CacheBuilder.newBuilder()
-                .expireAfterWrite(TimeUnit.MINUTES.toSeconds(1), TimeUnit.SECONDS)
+        this.inFlightRequests = CacheBuilder.newBuilder()
```
digitalbuddha: remove the `this` no reason for it

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -53,41 +53,60 @@
...
+                .expireAfterWrite(getCacheTTL(), getCacheTTLTimeUnit())
                 .build();
 
-        subject = BehaviorSubject.create();
+        this.subject = BehaviorSubject.create();
```
digitalbuddha: remove `this`

store/src/main/java/com/nytimes/android/external/store/base/impl/RealStoreBuilder.java
```diff
@@ -25,8 +25,9 @@
 public class RealStoreBuilder<Raw, Parsed, Key> {
     private final List<KeyParser> parsers = new ArrayList<>();
     private Persister<Raw, Key> persister;
-    private Cache<Key, Observable<Parsed>> memCache;
     private Fetcher<Raw, Key> fetcher;
+    private long expireAfter = 0;
+    private TimeUnit expireAfterTimeUnit = null;
```
brianPlummer: it might be a better approach to initialize this to a non-null value and set the expireAfter to a -1,  you can validate the input for memory() to guarantee non-nullness.  this would alleviate the need for null checking below.  another approach is to have the memory() method set a flag and use the passed in values for the builder (validate within memory() and just check flag below for inclusion...no need for validation..)

brianPlummer:
Thank you @pavlospt  for taking a stab at this!!! :)))))

store/src/main/java/com/nytimes/android/external/store/base/impl/RealStoreBuilder.java
pavlospt: @brianPlummer yes, that sounds great as well!

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
pavlospt: Will do!

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
pavlospt: Same here!

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
pavlospt: Sounds reasonable.

pavlospt:
@digitalbuddha @brianPlummer I added the changes we discussed. Let me know what you think. Tests are fine and seems to be working from testing on a real device as well!

pavlospt:
@digitalbuddha there is a remaining issue with PMD, that I don't really know how to fix. Any ideas? 
screen shot 2017-03-09 at 12 02 32

digitalbuddha:
Sadness. You can do a Pmd suppress warnings on the class. I guess we need
to refactor some stuff going forward.

app/src/main/java/com/nytimes/android/sample/SampleApp.java
```diff
@@ -57,11 +61,23 @@ private void initPersister() {
...
+                    @Nonnull
+                    @Override
+                    public Observable<RedditData> fetch(@Nonnull BarCode barCode) {
+                        Log.d("fetcher", "Going for fetcher");
+                        return provideRetrofit().fetchSubreddit(barCode.getKey(), "10");
+                    }
+                })
+//                .fetcher(barCode -> provideRetrofit().fetchSubreddit(barCode.getKey(), "10"))
+                .memoryPolicy(
+                    MemoryPolicy
```
digitalbuddha: Let's simplify this to MemoryPolicy.builder() which then returns an instance of the new builder.

digitalbuddha (commented):
Need to get this reviewed internally.

app/src/main/java/com/nytimes/android/sample/SampleApp.java
pavlospt: Ok will do in a bit!

digitalbuddha:
Will review in more detail today.

app/src/main/java/com/nytimes/android/sample/SampleApp.java
pavlospt: Done!

app/src/main/java/com/nytimes/android/sample/SampleApp.java
```diff
@@ -57,11 +61,22 @@ private void initPersister() {
...
-                .fetcher(barCode -> provideRetrofit().fetchSubreddit(barCode.getKey(), "10"))
-                .memory(CacheBuilder.newBuilder()
-                        .maximumSize(1)
-                        .expireAfterWrite(10, TimeUnit.SECONDS)
-                        .build())
+                .fetcher(new Fetcher<RedditData, BarCode>() {
+                    @Nonnull
+                    @Override
+                    public Observable<RedditData> fetch(@Nonnull BarCode barCode) {
+                        Log.d("fetcher", "Going for fetcher");
```
alexio: Remove log and switch back to lambda.

store/src/main/java/com/nytimes/android/external/store/base/impl/MemoryPolicy.java
```diff
@@ -0,0 +1,62 @@
+package com.nytimes.android.external.store.base.impl;
+
+
+import java.util.concurrent.TimeUnit;
+
+public class MemoryPolicy {
```
alexio: Since this is a public class, a javadoc would be appreciated. Could you write class/method comments where appropriate?

pavlospt: Ok will do!

app/src/main/java/com/nytimes/android/sample/SampleApp.java
pavlospt: Sure

alexio: Thanks! 👍

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -53,40 +53,79 @@
...
+        this.fetcher = fetcher;
+        this.persister = persister;
+        this.parser = parser;
+        this.stalePolicy = stalePolicy;
+
+        initMemCache(memoryPolicy);
+        initFlightRequests(memoryPolicy);
+
+        subject = BehaviorSubject.create();
     }
```
cybo42: Minor consideration, since this `init()` is almost the same as the one below, with the exception of creating the `defaultMemoryPolicy`, maybe consolidate the two, adding a conditional to configure a default `MemoryPolicy` if the none was passed in (or `null` passed in).  It may also make sense to move this into a constructor, since nothing else calls `init()`.

pavlospt: Agree, will fix that as well!

pavlospt:
@digitalbuddha @brianPlummer @alexio @cybo42 Finished the PR review changes. Let me know what you think!

alexio (approved):

brianPlummer (approved):

Issue #36, opened 2017-01-08
RxJava 2 support?


scottmeschke:
This is a pretty sweet little library, it's very similar to something I've been working on myself.
But RxJava 1 would makes me sad when inter-oping with the rest of my application code.

Any plans to update to RxJava 2?

The signifigant performance gains, Reactive Streams spec compliance, and Flowable vs Observable distinction alone makes it worth updating imo.

I would be willing to contribute or start a PR if there is interest.

digitalbuddha:
I was hoping to wait until some larger libraries like Retrofit make the plunge. Is it just  the public api you want to change or all rxjava1 to 2? I'm worried that current adoption of rx2 is low.  Will keep open

riclage:
Retrofit already supports RxJava 2 via an [adapter](https://github.com/JakeWharton/retrofit2-rxjava2-adapter).

My vote goes for changing everything to Rx 2 as keeping both dependencies increase the method count quite significantly. It could also be a solution that abstracts away the Rx dependency from this library altogether (e.g., using adapters).

digitalbuddha:
@scottmeschke I'd be open to figuring out a way to abstract RxJava stuff into adapters so that both Rx1 & Rx2 will work, I don't believe this would be trivial but at the same time possible.  Feel free to play with it (along with anyone else that wants rx2 support before we convert internally)

PaulWoitaschek:
I'd just make the transition to the latest version of RxJava and tell people with Lacey m legacy to use the interop library. 

Rx2 is the future.

pwittchen:
> tell people with old versions to use the interop library
Is there any interop solution, which allows converting RxJava2 to RxJava1 code? As far as I know [RxJava2Interop](https://github.com/akarnokd/RxJava2Interop) does this thing the other way around.

PaulWoitaschek:
Both ways @pwittchen

digitalbuddha:
We are making the jump to rx2 internally. Will give this a whirl soon. If anyone wants to convert feel free to start ☺️

PaulWoitaschek:
Does this go together with #22

digitalbuddha:
Yeah I guess its going to have to since some operators return single by default.. Good question!

PaulWoitaschek:
@digitalbuddha What about lambdas? With retrolamba it just compiles to bytecode like everything else and that would make the work way easier!

digitalbuddha:
[...]

PR #155, opened 2017-03-10
Port to RxJava 2

maksim-m:
Issue #36.

I started to port Store to RxJava 2.

- [x] Store
- [x] FileSystem
- [x] Sample app
- [x] Remove RxInterop dependency

### DIFF
--- a/app/src/main/java/com/nytimes/android/sample/activity/PersistingStoreActivity.java
+++ b/app/src/main/java/com/nytimes/android/sample/activity/PersistingStoreActivity.java
@@ -18,9 +18,11 @@ import com.nytimes.android.sample.reddit.PostAdapter;
 
 import java.util.List;
 
-import rx.Observable;
-import rx.android.schedulers.AndroidSchedulers;
-import rx.schedulers.Schedulers;
+
+import hu.akarnokd.rxjava.interop.RxJavaInterop;
+import io.reactivex.Observable;
+import io.reactivex.android.schedulers.AndroidSchedulers;
+import io.reactivex.schedulers.Schedulers;
 
 import static android.widget.Toast.makeText;
 
@@ -52,8 +54,10 @@ public class PersistingStoreActivity extends AppCompatActivity {
         }
     }
 
+    @SuppressWarnings("CheckReturnValue")
     public void loadPosts() {
         BarCode awwRequest = new BarCode(RedditData.class.getSimpleName(), "aww");
+
 
         this.persistedStore
                 .get(awwRequest)
@@ -74,8 +78,8 @@ public class PersistingStoreActivity extends AppCompatActivity {
     }
 
     private Observable<Post> sanitizeData(RedditData redditData) {
-        return Observable.from(redditData.data().children())
-                .map(Children::data);
+        return RxJavaInterop.toV2Observable(rx.Observable.from(redditData.data().children())
+                .map(Children::data));
     }
 
     @Override
--- a/app/src/main/java/com/nytimes/android/sample/activity/PersistingStoreActivity.java
+++ b/app/src/main/java/com/nytimes/android/sample/activity/PersistingStoreActivity.java
@@ -18,8 +18,6 @@ import com.nytimes.android.sample.reddit.PostAdapter;
 
 import java.util.List;
 
-
-import hu.akarnokd.rxjava.interop.RxJavaInterop;
 import io.reactivex.Observable;
 import io.reactivex.android.schedulers.AndroidSchedulers;
 import io.reactivex.schedulers.Schedulers;
@@ -77,8 +75,8 @@ public class PersistingStoreActivity extends AppCompatActivity {
     }
 
     private Observable<Post> sanitizeData(RedditData redditData) {
-        return RxJavaInterop.toV2Observable(rx.Observable.from(redditData.data().children())
-                .map(Children::data));
+        return Observable.fromIterable(redditData.data().children())
+                .map(Children::data);
     }
 
     @Override
--- a/app/src/main/java/com/nytimes/android/sample/activity/PersistingStoreActivity.java
+++ b/app/src/main/java/com/nytimes/android/sample/activity/PersistingStoreActivity.java
@@ -58,7 +58,6 @@ public class PersistingStoreActivity extends AppCompatActivity {
     public void loadPosts() {
         BarCode awwRequest = new BarCode(RedditData.class.getSimpleName(), "aww");
 
-
         this.persistedStore
                 .get(awwRequest)
                 .flatMap(this::sanitizeData)
--- a/filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
+++ b/filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
@@ -7,9 +7,10 @@ import java.io.FileNotFoundException;
 
 import javax.annotation.Nonnull;
 
+import hu.akarnokd.rxjava.interop.RxJavaInterop;
+import io.reactivex.Observable;
 import okio.BufferedSource;
 import rx.Emitter;
-import rx.Observable;
 import rx.functions.Action1;
 
 /**
@@ -31,12 +32,12 @@ public class FSReader<T> implements DiskRead<BufferedSource, T> {
     @Nonnull
     @Override
     public Observable<BufferedSource> read(@Nonnull final T key) {
-        return Observable.fromEmitter(new Action1<Emitter<BufferedSource>>() {
+        return RxJavaInterop.toV2Observable(rx.Observable.fromEmitter(new Action1<Emitter<BufferedSource>>() {
             @Override
             public void call(Emitter<BufferedSource> emitter) {
                 String resolvedKey = pathResolver.resolve(key);
                 boolean exists = fileSystem.exists(resolvedKey);
-              
+
                 if (exists) {
                     try {
                         BufferedSource bufferedSource = fileSystem.read(resolvedKey);
@@ -49,6 +50,6 @@ public class FSReader<T> implements DiskRead<BufferedSource, T> {
                     emitter.onCompleted();
                 }
             }
-        }, Emitter.BackpressureMode.NONE);
+        }, Emitter.BackpressureMode.NONE));
     }
 }
--- a/filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
+++ b/filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
@@ -7,11 +7,10 @@ import java.io.FileNotFoundException;
 
 import javax.annotation.Nonnull;
 
-import hu.akarnokd.rxjava.interop.RxJavaInterop;
 import io.reactivex.Observable;
+import io.reactivex.ObservableEmitter;
+import io.reactivex.ObservableOnSubscribe;
 import okio.BufferedSource;
-import rx.Emitter;
-import rx.functions.Action1;
 
 /**
  * FSReader is used when persisting from file system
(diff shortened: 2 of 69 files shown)


digitalbuddha:
Wow thank you for doing this! Will start reviewing. I don't think we can merge until we make a few changes internally like pulling in the interop but this was huge help!!

app/src/main/java/com/nytimes/android/sample/activity/PersistingStoreActivity.java
```diff
@@ -52,9 +54,11 @@ private void initStore() {
         }
     }
 
+    @SuppressWarnings("CheckReturnValue")
```
digitalbuddha, line 55: why this suppression on a void method?

app/src/main/java/com/nytimes/android/sample/activity/PersistingStoreActivity.java
```diff
@@ -52,9 +54,11 @@ private void initStore() {
...
     public void loadPosts() {
         BarCode awwRequest = new BarCode(RedditData.class.getSimpleName(), "aww");
 
+
```
digitalbuddha: nitpick: extra line

filesystem/src/test/java/com/nytimes/android/external/fs/StoreRefreshWhenStaleTest.java
```diff
@@ -69,7 +68,8 @@ public void diskWasRefreshedWhenStaleRecord() {
 
     }
 
-    @Test
+    // TODO
```
digitalbuddha: what happened here :-)

middleware-jackson/src/main/java/com/nytimes/android/external/store/middleware/jackson/JacksonReaderParser.java
```diff
@@ -30,7 +32,8 @@ public JacksonReaderParser(@Nonnull ObjectMapper objectMapper, @Nonnull Type typ
     }
 
     @Override
-    public Parsed call(@Nonnull Reader reader) {
+    @SuppressWarnings({"PMD.EmptyCatchBlock", "PMD.SignatureDeclareThrowsException"})
+    public Parsed apply(@NonNull Reader reader) throws Exception {
```
digitalbuddha: I think we should remove the try catch and now just throw the exception up.

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -186,15 +189,15 @@ public void call(Parsed parsed) {
     }
 
     void backfillCache(@Nonnull Key key) {
-        fetch(key).subscribe(new Action1<Parsed>() {
+        Disposable disposable = fetch(key).subscribe(new Consumer<Parsed>() {
```
digitalbuddha: not sure if it makes sense to make a disposable if nothing is going to dispose it

digitalbuddha (changes requested):
first pass, will have some more people look in am

app/src/main/java/com/nytimes/android/sample/activity/PersistingStoreActivity.java
maksim-m: `persistedStore.get.subscribe(...)` returns a `Disposable` object, which is unused. This causes [CheckReturnValue](http://errorprone.info/bugpattern/CheckReturnValue) error and the build fails.

filesystem/src/test/java/com/nytimes/android/external/fs/StoreRefreshWhenStaleTest.java
maksim-m: Work in progress :)

middleware-jackson/src/main/java/com/nytimes/android/external/store/middleware/jackson/JacksonStringParser.java
```diff
@@ -29,10 +30,10 @@ public JacksonStringParser(@Nonnull ObjectMapper objectMapper, @Nonnull Type typ
     }
 
     @Override
-    @Nullable
-    public Parsed call(@Nonnull String source) {
+    @SuppressWarnings({"PMD.EmptyCatchBlock"})
```
PaulWoitaschek: Why is nullable gone?

filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
```diff
@@ -31,12 +32,12 @@ public FSReader(FileSystem fileSystem, PathResolver<T> pathResolver) {
     @Nonnull
     @Override
     public Observable<BufferedSource> read(@Nonnull final T key) {
-        return Observable.fromEmitter(new Action1<Emitter<BufferedSource>>() {
+        return RxJavaInterop.toV2Observable(rx.Observable.fromEmitter(new Action1<Emitter<BufferedSource>>() {
```
PaulWoitaschek: I don't get it. Why is the interop library present at all?

PaulWoitaschek:
One thing to think about:

The fetcher implementations catch exceptions and return `null`. That will lead to a NPE when `map` is called.

filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
digitalbuddha: Hasn't converted that module in this PR :-)

digitalbuddha:
thanks for feedback @PaulWoitaschek I'll look into how to handle the null changes today.

middleware-jackson/src/main/java/com/nytimes/android/external/store/middleware/jackson/JacksonStringParser.java
maksim-m: Maybe we should use `Single` as the return type of parsers, to avoid possible NPE? (`Optional` is not available on Android)

digitalbuddha: My understanding of Rxjava2 is that nulls are not allowed anywhere. If source coming into parser is non null I feel that parser should be nonNull return as well.

digitalbuddha:
I think the fetcher should throw exception up rather than returning null.

middleware-jackson/src/main/java/com/nytimes/android/external/store/middleware/jackson/JacksonStringParser.java
PaulWoitaschek: Absolutely. What about adding a `throws Exception` to the parsers method signature and handling the errors from the caller?

digitalbuddha: yup makes sense. Store drops errors from disk read currently since there is network to fallback to.  If parsing error occurs from network it makes sense to throw error instead of overwriting the previous cached value with a null

maksim-m: Ok, I will fix it.
How about to use [ParserException](https://github.com/NYTimes/Store/blob/develop/store/src/main/java/com/nytimes/android/external/store/util/ParserException.java) instead of generic Exception, to have more specific error?

digitalbuddha:
Sounds good!

store/src/test/java/com/nytimes/android/external/store/KeyParserTest.java
```diff
@@ -49,8 +49,11 @@ public void testStoreWithKeyParserFuncNoPersister() throws Exception {
         testObservable.assertNoErrors()
                 .assertValues(NETWORK + KEY)
                 .awaitTerminalEvent();
-        // TODO assertTrue(testObservable.isDisposed()); ??
 
+        // This check doesn't work in RxJava 2.0.7
```
PaulWoitaschek: Its not necessary as `awaitTerminalEvent` ensures that the observer is in terminal state. And it's covered by RxJavas tests. So the comment can be removed completely.

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -130,7 +131,7 @@ private void init(Fetcher<Raw, Key> fetcher,
                 @Nonnull
                 @Override
                 @SuppressWarnings("PMD.SignatureDeclareThrowsException")
```
PaulWoitaschek: Warning can be removed

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -296,7 +297,7 @@ void notifySubscribers(Parsed data) {
     @Override
     public Observable<Parsed> stream(@Nonnull Key key) {
 
-        Observable<Parsed> stream = subject.asObservable();
+        Observable<Parsed> stream = subject;
```
PaulWoitaschek: call `hide()` to not expose the subject?

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -309,7 +310,7 @@ void notifySubscribers(Parsed data) {
     @Nonnull
     @Override
     public Observable<Parsed> stream() {
-        return subject.asObservable();
+        return subject;
```
PaulWoitaschek: hide()

store/src/test/java/com/nytimes/android/external/store/KeyParserTest.java
```diff
@@ -43,11 +45,15 @@ public String call(Integer integer, String s) {
...
+        TestObserver<String> testObservable = store.get(KEY).test().await();
         testObservable.assertNoErrors()
                 .assertValues(NETWORK + KEY)
-                .assertUnsubscribed();
+                .awaitTerminalEvent();
 
+        // This check doesn't work in RxJava 2.0.7
+        // See https://github.com/ReactiveX/RxJava/issues/4514
+        // TODO
+        // assertTrue(testObservable.isDisposed());
```
PaulWoitaschek: remove comment as `awaitTermianlEvent` ensures completion?

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
maksim-m: Good point. Thanks!

pavlospt:
@maksim-m Try rebasing remote/develop instead of pulling, to avoid `Merge ...` commits.

maksim-m:
Ok. Next time, I will!

digitalbuddha:
We have an internal meeting tomorrow where we will go over a plan to land this pr and what to do with branching/artifact names. Suggestions as always are welcome :-)

pavlospt:
@digitalbuddha One thing if I may. Please keep it on a separate release strategy and don't override the default modules that use RxJava1. :)

digitalbuddha:
I was thinking store2 package. Store will be deprecated, we will still push some updates until we switch completely internally

pavlospt:
@digitalbuddha Sounds fine!

digitalbuddha:
hi @maksim-m we got hit with a snowstorm for next 2 days. Will be discussing the plan to land this Thursday.  Sorry for keeping it open I promise we will get this in :-)

maksim-m:
No problem :)

digitalbuddha:
I'll be doing work internally in hopes of landing this Monday. The idea is to do same as rx and have store2 package rename so that someone can have store and store2 in same app until a full upgrade is made.

hwt88:
Hope to support RxJava 2 quickly .

digitalbuddha:
Working on it. Became not entirely simple due to us having to convert all
our parsers/fetchers/persisters that we us internally. Some had null apis.

digitalbuddha:
@maksim-m I created branch feature/rx2, do you mind changing base to that branch. What we want to do is merge your pr within the next few days. We will then change the packages to store2 within that branch and generate Store2:0.1.0-Snapshot artifacts on maven.  Once we get to a stable point and do the single/completable additions we will create rx1 branch freezing current master and then treat rx2 as the new master with a release of store2:1.0.0.  I think this will have the affect of letting whomever wants the rx2 features to use them shortly while not forcing everyone to switch over until we are sure the migration went well.

digitalbuddha (approved):

digitalbuddha:
Once checks pass I'll merge and fire off snapshots to maven we can then start thinking of making changes to `Single` to make @PaulWoitaschek happy :-D

maksim-m:
So, I changed the base branch to feature/rx2. The PR looks a bit weird now, but it should works.

digitalbuddha:
NP! I'm going to pull in and take over. Thanks for your big effort @maksim-m , if you ever come to NY stop by and I'll give you a tshirt at nytimes. :-)

maksim-m:
You are welcome! I'm glad I could help. :)

PR #166, opened 2017-03-21
Remove apt dependency and use annotationProcessor instead

ghost:
My commit removes the dependency to the apt plugin in the sample app and replaces it with the annotationProcessor from the android gradle plugin.


digitalbuddha:
Thank you for contribution!

PR #173, opened 2017-03-24
[WIP] read all for a particular barcode type

### LINKED ISSUE #172: Get all the items of a given type?
I'm currently only using the filesystem lib to persist data locally but I'm struggling to find a convenient way to fetch all items of a given type from my cache.

As far as I understood, currently I would need to keep another object in cache with the reference to objects I want returned and make sure I keep this parent object updated.

It seems this could be improved by making the "type" a folder in the filesystem by changing the `pathForBarcode()` method to something like:

```
    @Nonnull
    static String pathForBarcode(@Nonnull BarCode barCode) {
        return barCode.getType() + "\" + barCode.getKey();
    }

```

Then we could use the `Filesystem.list(barCode.getType())` method to get all the files from that folder and read them back to the requester.

Does that make sense? Is that something you would consider looking into? I could try to make a PR with that change.

digitalbuddha:
closes #172 
@riclage is this what you were thinking?

There is a breaking change in barcode resolution inside of SourceReader we were toStringing the barcode sometimes and other times we were appending key and type together. This PR unifies the behavior to be the same.

### DIFF
--- a/filesystem/src/main/java/com/nytimes/android/external/fs/BarCodePathResolver.java
+++ b/filesystem/src/main/java/com/nytimes/android/external/fs/BarCodePathResolver.java
@@ -7,7 +7,7 @@ import javax.annotation.Nonnull;
 class BarCodePathResolver implements PathResolver<BarCode> {
     @Nonnull
     @Override
-    public String resolve(@Nonnull BarCode key) {
-        return key.toString();
+    public String resolve(@Nonnull BarCode barCode) {
+        return barCode.getType()+"/"+barCode.getKey();
     }
 }
--- a/filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
+++ b/filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
@@ -10,6 +10,7 @@ import javax.annotation.Nonnull;
 import okio.BufferedSource;
 import rx.Emitter;
 import rx.Observable;
+import rx.exceptions.Exceptions;
 import rx.functions.Action1;
 import rx.functions.Func1;
 
@@ -63,7 +64,7 @@ public class FSReader<T> implements DiskRead<BufferedSource, T> {
                             try {
                                 return fileSystem.read(s);
                             } catch (FileNotFoundException e) {
-                                throw new RuntimeException(e);
+                                throw Exceptions.propagate(e);
                             }
                         }
                 });
--- a/filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
+++ b/filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
@@ -11,6 +11,7 @@ import okio.BufferedSource;
 import rx.Emitter;
 import rx.Observable;
 import rx.functions.Action1;
+import rx.functions.Func1;
 
 /**
  * FSReader is used when persisting from file system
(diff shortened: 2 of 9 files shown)


filesystem/src/main/java/com/nytimes/android/external/fs/BarCodePathResolver.java
```diff
@@ -7,7 +7,7 @@
 class BarCodePathResolver implements PathResolver<BarCode> {
     @Nonnull
     @Override
-    public String resolve(@Nonnull BarCode key) {
-        return key.toString();
+    public String resolve(@Nonnull BarCode barCode) {
+        return barCode.getType()+"/"+barCode.getKey();
```
riclage, line 11: You didn't want to keep using `toString()` and update the return value there?

filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
```diff
@@ -51,4 +52,20 @@ public void call(Emitter<BufferedSource> emitter) {
...
+    public Observable<BufferedSource>  readAll(@Nonnull final String type) throws FileNotFoundException {
+        return Observable
+                .from(fileSystem.list(type))
+                .map(new Func1<String, BufferedSource>() {
+                    @Override
+                    public BufferedSource call(String s) {
+                            try {
+                                return fileSystem.read(s);
+                            } catch (FileNotFoundException e) {
+                                throw new RuntimeException(e);
```
riclage: Why not `throw Exceptions.propagate(e);`?

riclage (commented):
That's it! Thank you very much.

I had just 2 small questions/suggestions that I left in the code. 

Also, for my case this fits perfectly but you didn't want to add the `readAll()` to the `Persister` interface?

riclage:
Another point is that this will invalidate existing cache files for apps using a previous version of the lib. Probably not a huge deal but something to be aware of.

digitalbuddha:
Yeah that was the whole breaking change thing. I'm going to sleep on it and see how I feel about this change.  Maybe this can be created as a new Persister with the other one deprecated for a few releases as warning.  Thanks for review!

filesystem/src/main/java/com/nytimes/android/external/fs/BarCodePathResolver.java
digitalbuddha: kept going back and forth on whether `toString` should have the "/" since that is only an implementation detail of the Reader.

digitalbuddha:
Didn't want to merge a breaking change, I'll PR in a new Persister that has this functionality and deprecate the current one.

riclage:
Let me know if there is anything I can do to help.

filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
```diff
@@ -51,4 +53,20 @@ public void call(Emitter<BufferedSource> emitter) {
             }
         }, Emitter.BackpressureMode.NONE);
     }
+
+
+    public Observable<BufferedSource>  readAll(@Nonnull final String type) throws FileNotFoundException {
+        return Observable
+                .from(fileSystem.list(type))
```
PaulWoitaschek, line 60: This should be deferred. Else when one doesn't immediately subscribe he will get a wrong result because `list` is called when he calls `readAll` and the actual read is performed when he subscribed (in the meantime there could be new items / old items deleted).

digitalbuddha: good call! running around at a few conference until next week. Will land and fix then.

PR #176, opened 2017-04-04
Close source after write to filesystem

alexio:


brianPlummer (approved):

cybo42 (approved):

filesystem/src/main/java/com/nytimes/android/external/fs/filesystem/FSFile.java
```diff
@@ -71,6 +71,9 @@ public void write(BufferedSource source) throws IOException {
             if (sink != null) {
                 sink.close();
             }
+            if (source != null) {
+                source.close();
+            }
```
cybo42: in future we may want to use the ` try-with-resources ` syntax to handle the `.close()`  and avoid need for the `finally`  block.
From what I read this only works on minSdk 19+ (not sure if that is a concern)

In this case we still want to delete the `tmpFile` so still need the `finally`

alexio: The minSdk bit would probably be cause for concern.

PaulWoitaschek: Why the null check? Makes more sense to me to annotate the source param as `@NonNull`. The whole method makes no sense with a null source and it should throw an exception if it's called with wrong arguments.

alexio: @PaulWoitaschek That's a fair point. Annotated the arg and removed the null check.

jessicachen07 (approved):

PR #178, opened 2017-04-06
Standardize store.stream() to emit only new items

ramonaharrison:
This PR proposes migrating the subject within `RealInternalStore` which acts as a bridge for subscribers to `store.stream()` from a `BehaviorSubject` to a `PublishSubject`. 

This is a go at standardizing the behavior of `store.stream()` so that subscribers will passively stream only new items -- that is, items that are fetched after they subscribe.

If I'm understanding right, in the current implementation, subscribers to `store.stream()` on a new `RealStore` instance will passively stream new items. However, subscribers to an existing store instance where the fetcher has already returned at least one item, in addition to passively streaming new items, will also receive the last fetched item as cached by the `BehaviorSubject`. As of now, this cached item also persists through a `clear()` event.

### DIFF
--- a/store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
+++ b/store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
@@ -23,7 +23,6 @@ import rx.functions.Action0;
 import rx.functions.Action1;
 import rx.functions.Func0;
 import rx.functions.Func1;
-import rx.subjects.BehaviorSubject;
 import rx.subjects.PublishSubject;
 
 import static com.nytimes.android.external.store.base.impl.StoreUtil.persisterIsStale;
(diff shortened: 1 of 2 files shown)


store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
```diff
@@ -313,15 +312,7 @@ void notifySubscribers(Parsed data) {
...
-
-        Observable<Parsed> stream = subject.asObservable();
-
-        //If nothing was emitted through the subject yet, start stream with get() value
-        if (!subject.hasValue()) {
-            return stream.startWith(get(key));
-        }
-
-        return stream;
+        return subject.asObservable().startWith(get(key));
```
PaulWoitaschek: No need for `asObservable()` as startWith is already an `Observable` and no longer a `Subject`

digitalbuddha (approved):
thanks for fixing!

store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
ramonaharrison: Great catch, thank you!

alexio (approved):

alexio:
Thanks for the fix, good work! :D

goldstei (approved):

cybo42 (approved):

Issue #22, opened 2017-01-05
Feature Request: Use Single<T> where appropriate


PaulWoitaschek:
I'm not sure but it seems to me that for certain operations there are more appropriate reactive base types that would make the API more fluently.

One candidate:
`Store.get(barcode : Barcode) : Observable<T>`
here a `Single<T>` would clearly indicate that its always just a single item that is being emitted.

digitalbuddha:
We went back and forth with this. 
Pro: more clear that only 1 emission
Con: Not all operators are implemented which limits how you can combine multiple stores into a single observable or dealing with long stacktraces when going `toObservable` and back a few times.   

We're not against it but to be honest I don't see much value in Observable over Single.  Maybe as part of the 2.x addition this can be added as well.

PaulWoitaschek:
RxJava 2 has more operators available and is better composable so that sounds like a good idea.

Also it would better match to the RxJava 2 API as its operators directly indicate the outcome too.
All its operators that return a single value return a Single, for example `.toList`.

digitalbuddha:
Something to think about. What happens if fetcher returns a stream? Single won't work for that case...

PaulWoitaschek:
That behavior is currently really undefined.

In response there is:

```java
fetcher()
                .fetch(key)
                .flatMap(new Func1<Raw, Observable<Parsed>>() {
                    @Override
                    public Observable<Parsed> call(Raw raw) {
                        return persister().write(key, raw)
                                .flatMap(new Func1<Boolean, Observable<Parsed>>() {
                                    @Override
                                    public Observable<Parsed> call(Boolean aBoolean) {
                                        return readDisk(key);
                                    }
                                });
                    }
                })
```

For flatMap the order of execution is not guaranteed and there will be a lot of read and writes concurrently without any synchronization. You might end up with `readDisk` results from previous `writeDisk` results or have written older responses first.

So either one handles that properly by using concatMap / switchMap and taking care of order of execution. Or you just define that it's a single value.

digitalbuddha:
Good point. What I was trying to figure out is how to make Store work with something like a websocket.  Back to brainstorming.

mradzinski:
[...]

Issue #181, opened 2017-04-16
networkBeforeStale() not calling onNext or onError


JoniDS:
When using **networkBeforeStale()** without the app having any cache (i.e. first run in flight mode), neither onNext() nor onError is called, instead onCompleted() is immediately called.

Expected outcome: onError() is called with the thrown exception (Retrofit connection exception in this case).
[StoreTest.zip](https://github.com/NYTimes/Store/files/924655/StoreTest.zip)

digitalbuddha:
Will look in am. Thanks for reporting that sounds not good ☺️

JoniDS:
I believe this is due to **FSReader.java** completing the observable (and ignoring the network exception) when the cache file doesn't exist:

```java
if (exists) {
    try {
        BufferedSource bufferedSource = fileSystem.read(resolvedKey);
        emitter.onNext(bufferedSource);
        emitter.onCompleted();
    } catch (FileNotFoundException e) {
        emitter.onError(e);
    }
} else { 
    ** emitter.onCompleted(); **
}
```

I have a working solution but need to check if I'm not breaking something else.
Basically this is what I'm doing:
FSReader.java:

```java
                if (exists) {
                    try {
                        BufferedSource bufferedSource = fileSystem.read(resolvedKey);
                        emitter.onNext(bufferedSource);
                        emitter.onCompleted();
                    } catch (FileNotFoundException e) {
                        emitter.onError(e);
                    }
                } else { 
                    ** emitter.onError(new Exception("File doesn't exist")); **
                    emitter.onCompleted();
                }
```

RealInternalStore.java
```java
    @Nonnull
    Observable<Parsed> response(@Nonnull final Key key) {
        return fetcher()
            .fetch(key)
            .flatMap(new Func1<Raw, Observable<Parsed>>() {
                @Override
                public Observable<Parsed> call(Raw raw) {
                    return persister().write(key, raw)
                        .flatMap(new Func1<Boolean, Observable<Parsed>>() {
                            @Override
                            public Observable<Parsed> call(Boolean aBoolean) {
                                return readDisk(key);
                            }
                        });
                }
            })
            .onErrorResumeNext(new Func1<Throwable, Observable<? extends Parsed>>() {
                @Override
                public Observable<? extends Parsed> call(Throwable throwable) {
                    if (stalePolicy == StalePolicy.NETWORK_BEFORE_STALE) {
                        **return readDisk(key, throwable);**
                    }
                    return Observable.error(throwable);
                }
[...]

PR #185, opened 2017-04-22
RxJava2 Migration: Return single/maybe where appropriate.

tairrzayev:
This attempts to deal with https://github.com/NYTimes/Store/issues/22

Most prominent changes:
- `Store` methods `get()` and `fetch()` will now return `Single`. 
- `Persister.read()` will now return `Maybe` because obviously it can complete without emissions if nothing was persisted before read.
 
This needs a good look at.


digitalbuddha (commented):
Wow thank you so much for contribution!  
We recently merged #181 on the rx1 side which fixed a bug using the `networkBeforeStale` directive.  The crux of the issue was that Disk returning empty prevented proper propagation of network errors on first launch.  I'll can spend some time tomorrow adding the necesary test to your branch and finding a solution that hopefully works with the `Maybe` type.  Here is the new test if you want to offer a suggestion. https://github.com/NYTimes/Store/blob/develop/filesystem/src/test/java/com/nytimes/android/external/fs/StoreNetworkBeforeStaleFailTest.java#L48

tairrzayev:
Thank you for heads up on that issue!

I went ahead and ported the original fix (https://github.com/NYTimes/Store/issues/181) into this branch in my latest commit - https://github.com/NYTimes/Store/pull/185/commits/20c0a2daf8864f247e7d57e81dbb617807ef01d7

tairrzayev:
I've noticed a couple of possible optimisations / improvements which could be made to the `FSReader.read()` in due course (which I think is a topic for a separate issue / discussion):


```
String resolvedKey = pathResolver.resolve(key);
boolean exists = fileSystem.exists(resolvedKey);

if (exists) {
    try {
        BufferedSource bufferedSource = fileSystem.read(resolvedKey);
        emitter.onSuccess(bufferedSource);
        emitter.onComplete();
    } catch (FileNotFoundException e) {
        emitter.onError(e);
    }
} else {
    emitter.onError(new FileNotFoundException(ERROR_MESSAGE + resolvedKey));
}
```

- `exists()` call seems redundant as `FileSystem.read()` already declares `FileNotFoundException` being thrown and in fact, it does the same check & throws it if the file does not exist in `FSFile.source()`, which backs the `FileSystem.read()` implementation.
- Arguably `emitter.onError()` from `else` branch leaks `FSReader` implementation details by returning the full path (which is derived in that method) while all the user should know if that there is no file matching their key.

With these points in mind, I'd propose the following implementation:

```
String resolvedKey = pathResolver.resolve(key);
try {
    BufferedSource bufferedSource = fileSystem.read(resolvedKey);
    emitter.onSuccess(bufferedSource);
    emitter.onComplete();
} catch (FileNotFoundException e) {
    emitter.onError(new FileNotFoundException("No file matching the key :" + key));
}
```

pavlospt:
@tairrzayev Sounds like a nice optimization and helps on the readability as well :)

digitalbuddha:
Thank you for pulling in fix! I'm still a bit concerned about Persister returning a maybe rather than a single. From the store's perspective a Persister  should return either a value or an error.  My concern is if a client implements their own persister and returns empty from it NetworkBeforeStale will not function correctly for the same reason that FSReader needed the pulled in fix.  Could you kindly make the adjustment to have Persister return a Single so that we can make sure that any user implemented Persister works correctly.

store/src/main/java/com/nytimes/android/external/store2/base/impl/RealInternalStore.java
```diff
@@ -171,24 +173,37 @@ private void initMemCache(MemoryPolicy memoryPolicy) {
...
 
-    Observable<Parsed> readDisk(@Nonnull final Key key) {
+    Maybe<Parsed> readDisk(@Nonnull final Key key) {
+        return readDisk(key, null);
+    }
+
+    Maybe<Parsed> readDisk(@Nonnull final Key key, final Throwable error) {
         return persister().read(key)
-                .onErrorResumeNext(new OnErrorResumeWithEmpty<Raw>())
+                .onErrorResumeNext(new Function<Throwable, MaybeSource<? extends Raw>>() {
```
digitalbuddha: Here's the part we need to either have Persisters return an error when  empty or have some other operator handle the empty maybe case when a throwable is passed into read disk.

tairrzayev:
Hmm, having looked back at the solution for this problem, I wonder if we could solve it as follows instead (keeping the `readDisk()` as-is, without adding additional `Throwable` param):

```
     @Nonnull
     Single<Parsed> response(@Nonnull final Key key) {
         return fetcher()
                 .fetch(key)
                 .flatMap(new Function<Raw, SingleSource<Parsed>>() {
                     @Override
                     public SingleSource<Parsed> apply(@NonNull Raw raw) {
                         return persister().write(key, raw)
                                 .flatMap(new Function<Boolean, SingleSource<Parsed>>() {
                                     @Override
                                     public SingleSource<Parsed> apply(@NonNull Boolean aBoolean) {
                                         return readDisk(key).toSingle();
                                     }
                                 });
                     }
                 })
                 .onErrorResumeNext(new Function<Throwable, SingleSource<? extends Parsed>>() {
                     @Override
                     public SingleSource<? extends Parsed> apply(@NonNull Throwable throwable) {
                         if (stalePolicy == StalePolicy.NETWORK_BEFORE_STALE) {
-                            return readDisk(key, throwable).toSingle();
+                            return readDisk(key)
+                                    .switchIfEmpty(Maybe.<Parsed>error(throwable))
+                                    .toSingle();
                         }
                         return Single.error(throwable);
                     }
                 })
                 .doOnSuccess(new Consumer<Parsed>() {
                     @Override
                     public void accept(@NonNull Parsed parsed) {
                         notifySubscribers(parsed);
                     }
                 })
                 .doAfterTerminate(new Action() {
                     @Override
                     public void run() {
                         inFlightRequests.invalidate(key);
                     }
                 })
                 .cache();
     }
```

In this case we'll switch to error if the cache is empty. It feels to me that reading from the cache should be a `Maybe` as it may or may not return the data - missing cached data should not really be an error.

digitalbuddha:
That looks lovely! I like your solution much better. Could you kindly make the adjustment. I will review this today and have a few review it tomorrow. Thank you again for your contribution!

tairrzayev:
Awesome :+1: - change applied.

ramonaharrison (approved):

alexio (approved):

digitalbuddha (approved):
Thank you for changes!

Issue #183, opened 2017-04-18
Add Lambdas


digitalbuddha:
Since the desugar tool will be a part of gradle 2.5 I think it's time to add retrolambda and start simplifying `RealInternalStore` among other classes.  

We are open to an external contribution for this.

pavlospt:
If anyone starts this, please mention it here, so that we can avoid double wasted work :)

Fitzpasd:
I'd like to tackle this.  The ask is to introduce RetroLambda in anticipation of moving to the desugar tool when that's ready?

PaulWoitaschek:
Make sure not to make use of try with resources as desugar does not support that yet but retrolambda does.

Fitzpasd:
Sure.  We've recently done this in work, only using the lamda part of RetroLambda (no try-with-resource, default methods, etc.)

digitalbuddha:
Sounds good please take it on. Don't forget about converting tests too please

ZacSweers:
I think the idea with desugar would be that you should write your library in java 8 and ship java 8 code, then desugar would run during the consuming app

digitalbuddha:
I believe we still need to convert a few modules before closing

PR #188, opened 2017-04-28
Add lambdas to Store and Filesystem modules

Fitzpasd:
Using Retrolambda to introduce lambdas and method references to the Store and Filesystem modules. 
 Issue #183.  Also increased the retrolambda gradle plugin to `3.6.1`


digitalbuddha (approved):
There is a 2x branch as well. I'd like to get that one converted at same time. I can do it this weekend unless you want to

Fitzpasd:
Sure, I'll write that one up also

brianPlummer (approved):

PR #130, opened 2017-02-13
Release 2.0.0

digitalbuddha:
Version 2.0.0 *(2017-02-13)*
----------------------------

**Breaking Changes**  - See Wiki/Closed PRs for more detail

* (#122) Store v1 removal in favor of Stores with generic keys - Store<ReturnType> becomes Store<ReturnType,Key>
* (#110) rework builders - See StoreBuilder for changes
* (#86)  Migrate Barcode to any Type - No longer need to use Barcode as your request type!

**New Features**
* (#94)  use javax annotations instead of intellij
* (#117) Feature/clear all cache - Disk Caching if Persister implements Clearable
* (#115) Feature/filepersister - Using Store with FileSystem no longer requires our BarCode type
* (#120) Avoid multiple resolves of the same Key
* (#111) networkBeforeStale and refreshOnStale - 2 ways to control Persisters that are StaleAware
* (#103) GetRefreshing - Like get but will repeat when clear
* (#113) lets try this travis snapshot deploy again - Snapshot deployment now works woo!

**Bug Fixes and Stability Improvements**
* (#125) Fix name on occurences of Key
* (#124) Create single instance of empty BarCode
* (#123) Simplify Persister clear and RefreshSubject notify 
* (#114) remove extranous exception throw
* (#108) Update to Moshi 1.4.0
* (#106) remove espresso, unneeded, conflicts with javax.annotations
* (#101) fixes inflight caching errors
* (#99)  Remove dead code and add/remove empty lines were needed
* (#97)  clear needs to clear noop disk and inflight
* (#126) Remove duplicate element from versions array

PR #194, opened 2017-05-10
Feature/all operations

sanchay-nyt:


digitalbuddha:
hey hey, I think you had a bad rebase, I don't believe there should have been more than a dozen files warranting a change.

filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
```diff
@@ -51,4 +53,21 @@ public void call(Emitter<BufferedSource> emitter) {
...
+    public Observable<BufferedSource> readAll(@Nonnull final T type) throws FileNotFoundException {
+        return Observable
+                .from(fileSystem.list(pathResolver.resolve(type)))
+                .map(new Func1<String, BufferedSource>() {
+                    @Override
+                    public BufferedSource call(String s) {
+                        try {
+                            return fileSystem.read(s);
+                        } catch (FileNotFoundException e) {
+                            throw Exceptions.propagate(e);
```
digitalbuddha, line 68: this is my Today I learned, neat!

store/src/main/java/com/nytimes/android/external/store/base/DiskErase.java
```diff
@@ -0,0 +1,19 @@
+package com.nytimes.android.external.store.base;
+
+import javax.annotation.Nonnull;
+
+import rx.Observable;
+
+public interface DiskErase<Raw, Key> {
```
digitalbuddha, line 7: Nice abstraction 💯

store/src/main/java/com/nytimes/android/external/store/util/NoopPersister.java
```diff
@@ -35,4 +37,10 @@
     public void clear(@Nonnull Key key) {
         networkResponses.remove(key);
     }
+
+    @Nonnull
+    @Override
+    public Observable<Raw> readAll(@Nonnull Key key) throws FileNotFoundException {
+        return null;
```
digitalbuddha, line 44: I'm not a fan of returning null here especially when Rxjava2 does not allow you to propogate a null to a consumer.  Ideally we can structure the base classes in a manner that would not need this override. If that is too messy then throw an exception from here instead.

digitalbuddha:
I think your rebase was off of rx2 branch and not develop.  I reviewed the actual meat of this pr and it looks good. The only thing that I think it is missing is a little better test coverage (I don't think there was a test for deleteAll).  Likewise there should be tests for the exception case as well.  Thank you for this PR it came together really nicely!

digitalbuddha:
looking at your commits it seems like #130 should not have been in your branch as this was from rx2 branch and not develop.

sanchay-nyt:
Thanks @digitalbuddha . With help from Brian, I have created a new PR https://github.com/NYTimes/Store/pull/207 . I will take care of the comments you posted today.

Issue #190, opened 2017-04-30
Add Raw to BufferedSource Converter


digitalbuddha:
Some users have asked for the ability to work with the `FileSystemPersister` when going from a raw that is already a Pojo or String for example a fetcher coming from `Retrofit`.  The initial proposal to create a `FileSystemPersister` which takes as input a raw to bufferedsouce converter ie:
```
userPojo->Okio.buffer(Okio.source(new ByteArrayInputStream(adapter.toJson(value)
      .getBytes(StandardCharsets.UTF_8))));
```

This way someone can configure a Store to take in a Pojo yet still work with our FileSystem.

saket:
I was just going to open an issue for this. I use a library that internally wraps in a set of APIs and does not expose their response bodies. This means that I can only use a `RealStoreBuilder<Parsed, Parsed, Key>` by using `StoreBuilder.key()`. Additionally, I want to use an object other than `Barcode` as the key because I want to have complex cache keys. Both these constraints mean that I will have to implement my own disk Persister.

In order to do that, I have been going through your code to understand what exactly are the pre-defined persisters like `SourcePersister` and `FileSystemPersister` are doing. So far I've realized that the filesystem module is tightly coupled with `Barcode`, which I feel is not right. Please correct me if I'm wrong.

The presence of a class named as `FileSystem` is also a small source of confusion, because Okio also uses a similar class with the same name. This cost me some time because I was trying to figure out how to use a BufferedSource, which you claim to be much faster.

Your Droidcon Italy talk convinced me to use Store, but I'm kinda struggling out in some places, partly because the library is low on documentation :)

digitalbuddha:
Filesystem is in no way ties to Barcodes.  Take a look at https://github.com/NYTimes/Store/blob/develop/filesystem/src/main/java/com/nytimes/android/external/fs/FileSystemPersisterFactory.java#L74 as this is a FileSystemPersister that can work with any key type.  The reason we use BufferedSource is that it allows streaming in a manner similar to Reader. As for naming, sorry nothing we can do about that now ☺️

Not sure if you had a question in there I did not answer.

saket:
Yes! Sadly `FileSystemPersister` is hardcoded to `BufferedSource`, which means I cannot use because I can only use `StoreBuilder.key()`. :( 

[...]

PR #198, opened 2017-05-12
Add Raw to BufferedSource transformer

ramonaharrison:
This PR adds an object to `BufferedSource` transformer per #190, so that users can configure a Store to take in a POJO (e.g. via a fetcher coming from Retrofit) and yet still work with the default `FileSystemPersister`.

Once everything looks good here, I can follow up with `TwoWayParserAdapter` implementations for Moshi and Jackson as well.

I'll create a seperate PR that updates the Sample app to demonstrate use of the transformer in the persisted version of the `RedditData` Store.


filesystem/src/main/java/com/nytimes/android/external/fs/BufferedSourceAdapter.java
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Converts Java values to JSON, and JSON values to Java.
+ */
+public interface BufferedSourceAdapter<Parsed> {
+    @Nonnull
+    BufferedSource toJson(@Nonnull Parsed value);
+
+    @Nonnull
+    Parsed fromJson(@Nonnull BufferedSource source);
```
leighsn: whats the `fromJson` for?

leighsn: fyi its mike, I'm on wife's computer and it was logged into her github :-)

app/src/main/java/com/nytimes/android/sample/SampleApp.java
```diff
@@ -79,17 +76,13 @@ private void initPersister() {
...
-    private BufferedSourceAdapter<RedditData> provideBufferedSourceAdapter() {
-        return new TwoWayParserAdapter<>(provideGson(), RedditData.class);
-    }
-
     private Persister<BufferedSource, BarCode> newPersister() throws IOException {
         return SourcePersisterFactory.create(getApplicationContext().getCacheDir());
     }
 
     private Observable<BufferedSource> fetcher(BarCode barCode) {
-        return provideRetrofit().fetchSubreddit(barCode.getKey(), "10").compose(new ObjectToSourceTransformer<>
```
digitalbuddha: I'm not crazy about the syntax that makes you first create a TwoWayParserAdapter followed by a ObjectToSourceTransformer, that feels like an implementation detail.  How about we make a `SourceFromObjectConverterFactory` so that we can do `SourceFromObjectConverterFactory.create(gson,Type)` which can actually return the `Transformer`.  Feel free to play with the name if something sounds better than SourceFromObjectConverterFactory, Spring was my first love, thus my lack of being able to create short names :-)

pavlospt: How about SourceProducer/SourceGenerator?

digitalbuddha: Oh I like SourceGenerator good call!

filesystem/src/main/java/com/nytimes/android/external/fs/BufferedSourceAdapter.java
ramonaharrison: I was going off the last comment on the issue when I added `fromJson` here, but I actually don't think it's needed since the Store's `Parser` implementation will handle this. I'll pull it out 👍

app/src/main/java/com/nytimes/android/sample/SampleApp.java
ramonaharrison: Good call, I'll update with these changes 👍

filesystem/src/test/java/com/nytimes/android/external/fs/ObjectToSourceTransformerTest.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    @Test
+    public void testTransformer() throws Exception {
+        BufferedSource source = Observable.just("test")
+                .compose(new ObjectToSourceTransformer<>(mockBufferedParser))
+                .toBlocking()
+                .first();
+
+        assertEquals(source, mockBufferedSource);
```
brianPlummer: assertJ please :)

digitalbuddha (approved):

digitalbuddha:
Could you mark these methods as @experimantal which would allow us to rename them without going up in major version number. Other than that I think this is good to go.

digitalbuddha:
I mean the factories

PR #199, opened 2017-05-14
expireAfterAccess added to MemoryPolicy

KevCron:
- I'm using the `expireAfterAccess` logic already in `CacheBuilder`, I'm assuming that's what the task was to do, only added some configuration to `MemoryPolicy`
- I didn't write any additional tests, as I couldn't find any that tested the `expireAfterWrite` logic of `MemoryPolicy` (let me know if I should add some tests)
- Not allowing for a client to set both `expireAfterWrite` and `expireAfterAccess`

### DIFF
--- a/cache/src/main/java/com/nytimes/android/external/cache/LocalCache.java
+++ b/cache/src/main/java/com/nytimes/android/external/cache/LocalCache.java
@@ -4760,7 +4760,7 @@ class LocalCache<K, V> extends AbstractMap<K, V> implements ConcurrentMap<K, V> 
     }
 
     /**
-     * Serializes the configuration of a LocalCache, reconsitituting it as an LoadingCache using
+     * Serializes the configuration of a LocalCache, reconstituting it as a LoadingCache using
      * CacheBuilder upon deserialization. An instance of this class is fit for use by the writeReplace
      * of LocalLoadingCache.
      * <p>
--- a/store/src/main/java/com/nytimes/android/external/store/base/impl/CacheFactory.java
+++ b/store/src/main/java/com/nytimes/android/external/store/base/impl/CacheFactory.java
@@ -20,18 +20,26 @@ final class CacheFactory {
                     .expireAfterWrite(StoreDefaults.getCacheTTL(), StoreDefaults.getCacheTTLTimeUnit())
                     .build();
         } else {
-            return CacheBuilder
-                    .newBuilder()
-                    .maximumSize(memoryPolicy.getMaxSize())
-                    .expireAfterWrite(memoryPolicy.getExpireAfter(), memoryPolicy.getExpireAfterTimeUnit())
-                    .build();
+            if (memoryPolicy.getExpireAfterAccess() == -1) {
+                return CacheBuilder
+                        .newBuilder()
+                        .maximumSize(memoryPolicy.getMaxSize())
+                        .expireAfterWrite(memoryPolicy.getExpireAfterWrite(), memoryPolicy.getExpireAfterTimeUnit())
+                        .build();
+            } else {
+                return CacheBuilder
+                        .newBuilder()
+                        .maximumSize(memoryPolicy.getMaxSize())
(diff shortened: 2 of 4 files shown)


store/src/main/java/com/nytimes/android/external/store/base/impl/MemoryPolicy.java
```diff
@@ -44,16 +50,28 @@ public long getMaxSize() {
     }
 
     public boolean isDefaultPolicy() {
-        return expireAfter == -1;
+        return expireAfterWrite == -1;
```
digitalbuddha: Let's make a constant that's equal to - 1 called DEFAULT_POLICY

cache/src/main/java/com/nytimes/android/external/cache/LocalCache.java
```diff
@@ -4760,7 +4760,7 @@ private Object readResolve() {
     }
 
     /**
-     * Serializes the configuration of a LocalCache, reconsitituting it as an LoadingCache using
+     * Serializes the configuration of a LocalCache, reconstituting it as a LoadingCache using
```
brianPlummer, line 4763: thanks!!!

store/src/main/java/com/nytimes/android/external/store/base/impl/CacheFactory.java
```diff
@@ -20,18 +20,26 @@ private CacheFactory() {
                     .expireAfterWrite(StoreDefaults.getCacheTTL(), StoreDefaults.getCacheTTLTimeUnit())
                     .build();
         } else {
-            return CacheBuilder
-                    .newBuilder()
-                    .maximumSize(memoryPolicy.getMaxSize())
-                    .expireAfterWrite(memoryPolicy.getExpireAfter(), memoryPolicy.getExpireAfterTimeUnit())
-                    .build();
+            if (memoryPolicy.getExpireAfterAccess() == -1) {
```
brianPlummer: what's -1?? is it the DEFAULT_POLICY? if so...maybe reference this value so it is clear :)

brianPlummer:
great work!! thank you!!

store/src/main/java/com/nytimes/android/external/store/base/impl/MemoryPolicy.java
```diff
@@ -44,16 +51,28 @@ public long getMaxSize() {
...
-        private long expireAfter = -1;
+        private long expireAfterWrite = -1;
+        private long expireAfterAccess = -1;
         private TimeUnit expireAfterTimeUnit = TimeUnit.SECONDS;
         private long maxSize = 1;
 
-        public MemoryPolicyBuilder setExpireAfter(long expireAfter) {
-            this.expireAfter = expireAfter;
+        public MemoryPolicyBuilder setExpireAfterWrite(long expireAfterWrite) {
+            if (expireAfterAccess != -1) {
```
digitalbuddha: Use constant here too, there should be no references to - 1 besides the constant

digitalbuddha (commented):
For more -1s and then lgtm

store/src/main/java/com/nytimes/android/external/store/base/impl/MemoryPolicy.java
KevCron: Ok

digitalbuddha:
As for tests, tests are always good especially for parts that have logic. There's a few conditionals and comparisons which ideally would have tests to verify that you didn't accidently reverse a comparison as well as to prevent future regressions

KevCron:
Makes sense. Should I write some additional tests then, or are there more important things to work on?

digitalbuddha:
Test would be awesome. Thank you so much! I'm in Europe and going to bed,
will check over in am if you get it done tonight.  I really appreciate your
contributions lately.

ramonaharrison:
This looks great! I'll merge this PR now so that it can be included in the snapshot for the next release. Feel free to follow up with any tests you're working on in a separate PR.

PR #206, opened 2017-05-16
update sample app to use maven version of lib

brianPlummer:


pavlospt:
Shouldn't it be pointing to 2.0.4?

brianPlummer:
@pavlospt oh no!!! i messed that up...i will close this one and re-do it...thank you for catching this..

PR #207, opened 2017-05-16
ReadAll/ClearAll

sanchay-nyt:


filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
```diff
@@ -32,29 +35,41 @@ public FSReader(FileSystem fileSystem, PathResolver<T> pathResolver) {
...
-            } else {
-                emitter.onError(new FileNotFoundException(ERROR_MESSAGE + resolvedKey));
             }
         }, Emitter.BackpressureMode.NONE);
     }
+
+    @Nonnull
+    @Override
+    public Observable<BufferedSource> readAll(@Nonnull final T type) throws FileNotFoundException {
+        return Observable
```
digitalbuddha: Pls wrap the observable.from in observable.defer

digitalbuddha (commented):
1 minor issue which keeps the readall from being lazy. Other than that lgtm

filesystem/src/main/java/com/nytimes/android/external/fs/FSReader.java
sanchay-nyt: Made changes. Is this what you had in mind @digitalbuddha?

digitalbuddha:
Yup that it! Last part, run the gradle task check locally to see and fix static analyzer violations. I think a variable has wrong case.

sanchay-nyt:
Got it! Looking into it. I am also writing test cases for deleteAll.

Thanks,
Sanchay

store/src/main/java/com/nytimes/android/external/store/base/AllPersister.java
```diff
@@ -0,0 +1,24 @@
...
+     *                throw an exception
+     */
+    @Nonnull
+    Observable<Raw> readAll(@Nonnull final String path) throws FileNotFoundException;
+
+    /**
+     * @param path to delete all the date in the the path.
+     */
+    @Nonnull
+    Observable<Boolean> deleteAll(@Nonnull final String path);
```
ramonaharrison: What's the difference between `readAll(path)` and `deleteAll(path)` here versus in the `DiskAllErase` and `DiskAllRead` interfaces? Could `AllPersister` just implement those interfaces, similar to how the current `Persister` works?

filesystem/src/main/java/com/nytimes/android/external/fs/SourcePersister.java
```diff
@@ -54,4 +51,9 @@ static String pathForBarcode(@Nonnull BarCode barCode) {
         return sourceFileWriter.write(barCode, data);
     }
 
+    @Nonnull
+    static String pathForBarcode(@Nonnull BarCode barCode) {
+        return barCode.toString();
```
ramonaharrison: Will this change in path resolution break anything for current users of `SourcePersister`?

filesystem/src/main/java/com/nytimes/android/external/fs/SourceFileAllReader.java
```diff
@@ -0,0 +1,13 @@
+package com.nytimes.android.external.fs;
+
+import com.nytimes.android.external.fs.filesystem.FileSystem;
+import com.nytimes.android.external.store.base.DiskAllErase;
+import com.nytimes.android.external.store.base.DiskAllRead;
+
+
+public class SourceFileAllReader extends FSAllReader implements DiskAllRead {
```
ramonaharrison: How is this class used? Why not have `SourceAllPersister` encapsulate `FSAllReader` directly? (same question goes for `SourceFileEraser` + `SourceFileAllEraser`)

filesystem/src/test/java/com/nytimes/android/external/fs/FSReaderTest.java
```diff
@@ -0,0 +1,54 @@
...
+        BarCode barCode1 = new BarCode(TYPE, "key2.txt");
+
+        File tempDir = createTempDir();
+        String str = tempDir.getAbsolutePath();
+        System.out.println(str);
+        FileSystem fileSystem = FileSystemFactory.create(tempDir);
+
+        //write different data to File System for each barcode
+        fileSystem.write(barCodeWritePathResolver.resolve(barCode), source(CHALLAH));
+        fileSystem.write(barCodeWritePathResolver.resolve(barCode1), source(CHALLAH_CHALLAH));
```
ramonaharrison: Use `BarCodeReadAllPathResolver` for these writes

filesystem/src/test/java/com/nytimes/android/external/fs/FSReaderTest.java
```diff
@@ -0,0 +1,54 @@
...
+        FSReader<BarCode> reader = new FSReader<>(fileSystem, barCodePathResolver);
```
ramonaharrison: Should be `FSAllReader` and `readAll` below

filesystem/src/test/java/com/nytimes/android/external/fs/FSReaderTest.java
```diff
@@ -0,0 +1,54 @@
...
+        //read back all values for the TYPE
+        BlockingObservable<BufferedSource> observable = reader.read(barCode).toBlocking();
+        assertThat(observable.first().readUtf8()).isEqualTo(CHALLAH);
+        assertThat(observable.last().readUtf8()).isEqualTo(CHALLAH_CHALLAH);
```
ramonaharrison: I just tested this out and had to reverse these asserts -- `CHALLAH` will actually be last while `CHALLAH_CHALLAH` will be first because of the ordering of items returned by `filesystem.list()`. 

Since order doesn't matter here, maybe call `toList()` on this Observable and then just assert that it contains the expected items and is of the expected size?

ramonaharrison (commented):
These changes look great! I've added a few suggestions in line. Thank you so much for all of your work on this so far!

store/src/main/java/com/nytimes/android/external/store/base/AllPersister.java
sanchay-nyt: Yup. Made the change

filesystem/src/main/java/com/nytimes/android/external/fs/SourcePersister.java
sanchay-nyt: It's not my change. For some reason this method is showing up a few lines below. I have reverted it back to how it is in develop branch

filesystem/src/main/java/com/nytimes/android/external/fs/SourceFileAllReader.java
sanchay-nyt: Yes, you are right. I built it like SourceFileReader which has getRecordState method. But such a method does not make sense in AllOperations. Deleting these file.

filesystem/src/test/java/com/nytimes/android/external/fs/FSReaderTest.java
sanchay-nyt: Well, I did not change this test from our last debugging session since I was not sure if I was on right track. Working on this now. :)

sanchay-nyt:
@ramonaharrison Please take a look. Do we need BarCodePathAllResolver anymore now?

store/src/main/java/com/nytimes/android/external/store/base/AllPersister.java
```diff
@@ -0,0 +1,27 @@
...
+
+
+import java.io.FileNotFoundException;
+
+import javax.annotation.Nonnull;
+import javax.annotation.OverridingMethodsMustInvokeSuper;
+
+import rx.Observable;
+
+public interface AllPersister<Raw> extends DiskAllRead, DiskAllErase {
```
ramonaharrison: We should also implement `Persister` here, so that users can pass an `AllPersister` to `StoreBuilder.persister()` and access the traditional `read()` and `write()` methods in addition to these new all methods.

Then in `SourceAllPersister`, add an FSReader and FSWriter that will use your new `BarCodeReadAllPathResolver` for those `read(barcode)` and `write(barcode)` operations (similar to how `FileSystemPersister` works).

ramonaharrison (commented):
Awesome, these changes look great! One comment in line about implementing `Persister` and where I think the `BarCodeReadAllPathResolver` fits in.

Otherwise, just fix the PMD stuff and lgtm!

sanchay-nyt:
@ramonaharrison  I have made changes. Should I write SourceAllPersisterTest just like SourcePersisterTest? I am working on PMD..unfortunately gradlew check is not working in my local.

ramonaharrison:
This looks great @sanchay-nyt. Adding tests for SourceAllPersister would be much appreciated. Otherwise, once one more reviewer gets a chance to take a look, I think it's good to go.

Thank you!

digitalbuddha:
Example on Readme or wiki on how to use it would be optimal as well.

sanchay-nyt:
@ramonaharrison Sure. Will write test cases for all persister .
@digitalbuddha  Sure. Will update Readme.

riclage:
@digitalbuddha Could this PR be used already or do you think there will be breaking changes?

digitalbuddha (approved):

digitalbuddha:
there will be no breaking changes.  You will need to use the new Resolver that is in this pr.  Additionally you will need to use the new Persister. By doing so you will now be able to call persister.readAll(type) or persister.deleteAll(type) directly to clear all barcodes for a particular type.

digitalbuddha:
this is last feature we are merging into rx1 branch prior to deprecating it.

PR #208, opened 2017-05-16
correct sample to use maven version of lib

brianPlummer:


brianPlummer:
@pavlospt here is the correct change....thank you again!! :))

ramonaharrison (approved):

PR #218, opened 2017-05-22
Update to RxJava 1.3.0 and use stable create(Action)

ZacSweers:
`Observable.fromEmitter` was removed in 1.3.0.


ZacSweers:
In general the rule of thumb I've seen is to not use experimental/beta APIs in libraries for stability, and was burned by this when I updated my app's RxJava version :/

alexio (approved):

ramonaharrison (approved):

PR #229, opened 2017-06-13
Add store-kotlin module

stoyicker:
This PR adds a new artifact, store-kotlin, which compiles regular Store and includes a few handy methods for Store initialization when using Kotlin.

I doubt this will seamlessly plug onto the current release procedure because the documentation generated by dokka goes somewhere else than the traditional javadoc by default, and I'm not sure what approach to take - could probably use some help with regards to that.
Also, a few tests use reflection via commons lang `EqualsBuilder.reflectionEquals(...)`. If in the future the relevant attributes gain more relaxed access it may be a good idea to edit the tests.

### DIFF
--- a/store-kotlin/src/main/AndroidManifest.xml
+++ b/store-kotlin/src/main/AndroidManifest.xml
@@ -1,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest
+    package="org.jorge.store.kotlin"/>
--- a/store-kotlin/src/main/AndroidManifest.xml
+++ b/store-kotlin/src/main/AndroidManifest.xml
@@ -1,3 +1,3 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest
-    package="org.jorge.store.kotlin"/>
+    package="com.nytimes.android.external.store-kotlin"/>
--- a/store-kotlin/src/main/kotlin/com/nytimes/android/external/store3/base/impl/FluentMemoryPolicyBuilder.kt
+++ b/store-kotlin/src/main/kotlin/com/nytimes/android/external/store3/base/impl/FluentMemoryPolicyBuilder.kt
@@ -1,0 +1,28 @@
+package com.nytimes.android.external.store3.base.impl
+
+/**
+ * Wraps methods for fluent MemoryPolicy instantiation.
+ */
+class FluentMemoryPolicyBuilder private constructor() {
+    companion object {
+        /**
+         * Builds a MemoryPolicy.
+         * @param config The desired configuration for the memory policy.
+         * @return A MemoryPolicy with the desired parameters or default ones if none were provided.
+         */
+        fun build(config: (MemoryPolicyParameters.() -> Unit)? = null): MemoryPolicy =
+            MemoryPolicyParameters().apply { config?.invoke(this) }.let {
+                MemoryPolicy.builder()
+                        .apply {
+                            if (it.expireAfterAccess != MemoryPolicy.DEFAULT_POLICY) {
+                                this.setExpireAfterAccess(it.expireAfterAccess)
+                            } else if (it.expireAfterWrite != MemoryPolicy.DEFAULT_POLICY) {
+                                this.setExpireAfterWrite(it.expireAfterWrite)
+                            }
+                        }
+                        .setExpireAfterTimeUnit(it.expireAfterTimeUnit)
+                        .setMemorySize(it.memorySize)
+                        .build()
+            }
+    }
+}
--- a/store-kotlin/src/main/kotlin/com/nytimes/android/external/store3/base/impl/FluentMemoryPolicyBuilder.kt
+++ b/store-kotlin/src/main/kotlin/com/nytimes/android/external/store3/base/impl/FluentMemoryPolicyBuilder.kt
@@ -1,8 +1,11 @@
 package com.nytimes.android.external.store3.base.impl
 
+import com.nytimes.android.external.store3.annotations.Experimental
+
 /**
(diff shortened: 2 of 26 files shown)


store-kotlin/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest
+    package="org.jorge.store.kotlin"/>
```
pavlospt: @stoyicker is this one correct?

store-kotlin/src/test/kotlin/com/nytimes/android/external/store3/base/impl/MemoryPolicy.kt
```diff
@@ -0,0 +1,9 @@
+package com.nytimes.android.external.store3.base.impl
+
+//when(other) {
```
pavlospt: @stoyicker Is this file needed, since it is fully commented?

pavlospt (commented):
Great work! One comment, we could introduce a better DSL for MemoryPolicy I guess since Kotlin allows us to do so :) something like: 
```kotlin
memoryPolicy {
    expireAfterWrite = ...
    expireAfterAccess = ...
    expireAfterTimeUnit = ...
    memorySize = ...
}
```

Let me know what you think :)

digitalbuddha:
your base branch was off I changed it to `rx2`.  As for apache commons I'd prefer for you to take out that dependency as it is an additional dependency we need to manage in future (for only 1 line).  Could you also switch to using assertJ fluent assertions.  Not using dokka, you can remove that as well. I'll do a full review tonight. Thank you so much for contribution!!

store-kotlin/src/main/AndroidManifest.xml
stoyicker: ooopsie

stoyicker:
First off, thank you guys, always a pleasure to help the project :)
@pavlospt I already included a dsl for MemoryPolicy which works exactly like what you describe, it just needs to be accessed through `FluentMemoryPolicyBuilder`. The reason for this is consistency: the new builder for Store is called `FluentStoreBuilder` because the methods in `StoreBuilder` are static so I can't add their Kotlin parts as extensions functions since static extension functions are not supported. Therefore I made the dsl for `MemoryPolicy` work the same way, even if _technically_ I could use an extension function for it since `MemoryPolicy` is not built through static methods but using a builder instance instead.
Now @digitalbuddha your things:
- Bring in commons is bad, I agree, especially when it's used so little. But I need a way to make value-based comparisons in the tests, and I don't think adding the code for reflection-based comparison myself is a good idea either. What do you think is a good approach here?
- As far as I know Dokka is not optional, I don't think the Kotlin plugin uses Javadoc at all since running the task on the store-kotlin module does nothing, so removing Dokka is removing all documentation for the module :/

digitalbuddha:
I think you had a bad rebase, says there are 176 files touched and other commits within the branch.

pavlospt:
@digitalbuddha if the branch that he checked out from to start this was not the Rx2 one, I think the rebase is legit, since it contains all the package name changes.

stoyicker:
everything wrt to the base change has been done by github, so now that you guys mention I should probably rebase to drop the commits from the other branch so that the diff correctly matches what this pr is about. will do later today

pavlospt:
@stoyicker Impressive work, huge congrats for this one!

stoyicker:
thanks :3
I've added a bit of a 'hack' to include kdoc in the javadoc jar so it doesn't go empty to mavencentral, even though it won't be parsable by IDEs. With that, everything other than removing commons lang should now be addressed I think, we just need to decide on that.

digitalbuddha:
thank you, will rereview tonight

brianPlummer:
LGTM! great work!

brianPlummer (approved):

store-kotlin/src/main/kotlin/com/nytimes/android/external/store3/base/impl/FluentMemoryPolicyBuilder.kt
```diff
@@ -0,0 +1,28 @@
+package com.nytimes.android.external.store3.base.impl
+
+/**
+ * Wraps methods for fluent MemoryPolicy instantiation.
+ */
+class FluentMemoryPolicyBuilder private constructor() {
```
digitalbuddha: lets mark this class with @Experimental annotation. you can create that as a new marker interface. This way we can take a few minor revisions to stabilize api prior to shipping

store-kotlin/src/main/kotlin/com/nytimes/android/external/store3/base/impl/FluentRealStoreBuilder.kt
```diff
@@ -0,0 +1,46 @@
...
+import com.nytimes.android.external.store3.base.Fetcher
+import com.nytimes.android.external.store3.base.Parser
+import com.nytimes.android.external.store3.base.Persister
+import com.nytimes.android.external.store3.util.KeyParser
+
+/**
+ * A fluent builder for Store instantiation.
+ * @param fetcher The fetcher for the created instance.
+ */
+internal class FluentRealStoreBuilder<Raw, Parsed, Key> constructor(
```
digitalbuddha, line 12: same mark as @Experimental (anything that is public api)

digitalbuddha (commented):
few annotations and then lgtm

stoyicker:
thanks! I've added an Experimental annotation of our own and I have also updated usages of rx's one to it. As for the new artifact, all public classes are now annotated as Experimental.

Issue #226, opened 2017-06-08
Distinguish the source of data that a Store feeds


pavlospt:
So the other day we were discussing with @charbgr that it would be great to know where the data from a `Store`, is coming from. It can be quite helpful in specific UI updates when loading data, e.g.: let a user know that the data she/he sees are possibly outdated :)

We suggest of adding a wrapper class that could distinguish the source for the data it holds. 

This could be an `Enum` or an `IntDef` whatever you prefer most. 

Let us know what you think about it and whether you need us to start a PR or something to help if you like the idea :)

digitalbuddha:
I'm not sure this needs to be feature if store. You could always create your own Response object and set a field on it in your fetcher or persister.

pavlospt:
@digitalbuddha If this is included in Store it will be 100% that you will always know the source of data, instead of having to duplicate code in all your possible fetchers and persisters :)

digitalbuddha:
#227 also wants wrapping to emit state. Maybe we can take both requests up on the rx2 branch prior to first real release of it.

digitalbuddha:
close as dupe of #227

PR #250, opened 2017-07-21
Return object with meta data

charbgr:
Hey,

this is an attempt to return an object with meta data.

I introduced a Result model which returns the parsed object accompanied by its source(network/cache). I couldn't distinguish if the cache is from memory or disk due to the fact that memory works as an "interceptor".

Let me know what you think.


digitalbuddha:
closes https://github.com/NYTimes/Store/issues/226

build.gradle
```diff
@@ -86,11 +86,11 @@ subprojects {
             project.android.dexOptions.preDexLibraries = rootProject.ext.preDexLibs
         }
     }
-    project.plugins.apply('net.ltgt.errorprone')
-
-    configurations.errorprone {
-        resolutionStrategy.force 'com.google.errorprone:error_prone_core:2.0.15'
-    }
+//    project.plugins.apply('net.ltgt.errorprone')
```
digitalbuddha: ?

store/src/main/java/com/nytimes/android/external/store/util/Result.java
```diff
@@ -0,0 +1,102 @@
...
+     * Convenient method to create a result object from {SOURCE_NETWORK}.
+     * @param value data type after parsing
+     * @param <T> data type after parsing
+     * @return a Result object with T passed
+     */
+    public static <T> Result<T> createFromNetwork(T value) {
+        return new Result<>(SOURCE_NETWORK, value);
+    }
+
+    public String getSource() {
```
digitalbuddha: remove get from name pls (unless there is also a set)

build.gradle
pavlospt: @digitalbuddha That's probably a mistake, since the project wouldn't run with error-prone enabled.

store/src/main/java/com/nytimes/android/external/store/util/Result.java
charbgr: Done! :)

build.gradle
charbgr: I rebased it and remove the comments. Sorry for the confusion!

store/src/main/java/com/nytimes/android/external/store/util/Result.java
```diff
@@ -0,0 +1,102 @@
...
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        Result<?> result = (Result<?>) o;
+
+        if (source == null) {
```
wdziemia: if `this.source` is null and o.source is not null, return that they are equal?

store/src/main/java/com/nytimes/android/external/store/util/Result.java
```diff
@@ -0,0 +1,102 @@
...
+            return result.source != null;
+        } else {
+            if (!source.equals(result.source)) {
+                return false;
+            }
+        }
+
+        if (value == null) {
```
wdziemia: if `this.value` is null and o.value is not null, return that they are equal?

store/src/main/java/com/nytimes/android/external/store/util/Result.java
```diff
@@ -0,0 +1,102 @@
...
+            return result.value != null;
+        } else {
```
wdziemia: can be simplified to `else { return value.equals(result.value) }`

store/src/main/java/com/nytimes/android/external/store/util/Result.java
```diff
@@ -0,0 +1,102 @@
...
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        Result<?> result = (Result<?>) o;
+
+        if (source == null) {
+            return result.source != null;
+        } else {
+            if (!source.equals(result.source)) {
```
wdziemia: can be simplified to `else { return source value.equals(source.value) }`

wdziemia (changes requested):

store/src/main/java/com/nytimes/android/external/store/util/Result.java
charbgr: @wdziemia what do you mean by `else { *source* value.equals(source.value) }`

wdziemia: @charbgr whoops, edited the original

cybo42:
(nitpick) remove the `.DS_Store` and `app/.DS_Store` files from the PR

store/src/main/java/com/nytimes/android/external/store/util/Result.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Result is a container with meta data for the parsed object.
+ *
+ * @param <Parsed> data type after parsing
+ */
+public final class Result<Parsed> {
+
+    public static final String SOURCE_CACHE = "disk";
+    public static final String SOURCE_NETWORK = "network";
+
```
cybo42: Any thoughts on using an Enum to identify source, something like below?
```
   public static enum Source {
        DISK, NETWORK
    }
```

charbgr: @cybo42 I am ok with that! I didn't do it due to increase in DEX.

store/src/main/java/com/nytimes/android/external/store/util/Result.java
charbgr: Fixed

store/src/main/java/com/nytimes/android/external/store/util/Result.java
charbgr: Fixed

store/src/main/java/com/nytimes/android/external/store/util/Result.java
charbgr: Done

digitalbuddha:
lgtm

cybo42 (approved):

digitalbuddha:
Thank you for contribution! Snapshot will be live in twenty minutes. We will make a store 3.0 release this week.

PR #252, opened 2017-07-24
fix stream for a single barcode

### LINKED ISSUE #112: Fix store.stream(key)
I'd like to undeprecate Observable<T> stream(V id).  The adjusted functionality will be an endless stream that emits new network values coming through a store for a particular barcode. Stream allows you to stay subscribed in many places and then calling fetch(key)  to get them all a new value. Implementation will be similar to getRefreshing but will instead hook into new network values.

The difference between getRefreshing(key) & stream(key) is that getRefreshing will repeat when clear is called while stream will repeat when anything hits the fetcher again.

digitalbuddha:
This pr fixes the functionality of stream(key) which will now kick off a get request and then propogate a new network value anytime it comes through the store for the same key.

closes #112

### DIFF
--- a/filesystem/src/main/java/com/nytimes/android/external/fs3/SourcePersisterFactory.java
+++ b/filesystem/src/main/java/com/nytimes/android/external/fs3/SourcePersisterFactory.java
@@ -77,7 +77,4 @@ public final class SourcePersisterFactory {
         }
         return SourcePersister.create(fileSystem);
     }
-
-
-
 }
--- a/filesystem/src/main/java/com/nytimes/android/external/fs3/SourcePersisterFactory.java
+++ b/filesystem/src/main/java/com/nytimes/android/external/fs3/SourcePersisterFactory.java
@@ -78,4 +78,6 @@ public final class SourcePersisterFactory {
         return SourcePersister.create(fileSystem);
     }
 
+
+
 }
--- a/store/src/main/java/com/nytimes/android/external/store3/base/impl/RealInternalStore.java
+++ b/store/src/main/java/com/nytimes/android/external/store3/base/impl/RealInternalStore.java
@@ -6,6 +6,7 @@ import com.nytimes.android.external.store3.base.InternalStore;
 import com.nytimes.android.external.store3.base.Persister;
 import com.nytimes.android.external.store3.util.KeyParser;
 
+import java.util.AbstractMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
 
@@ -35,7 +36,7 @@ final class RealInternalStore<Raw, Parsed, Key> implements InternalStore<Parsed,
 
     private final PublishSubject<Key> refreshSubject = PublishSubject.create();
     private Fetcher<Raw, Key> fetcher;
-    private PublishSubject<Parsed> subject;
+    private PublishSubject<AbstractMap.SimpleEntry<Key, Parsed>> subject;
 
     RealInternalStore(Fetcher<Raw, Key> fetcher,
                       Persister<Raw, Key> persister,
@@ -79,7 +80,7 @@ final class RealInternalStore<Raw, Parsed, Key> implements InternalStore<Parsed,
     public Observable<Parsed> getRefreshing(@Nonnull final Key key) {
         return get(key)
                 .toObservable()
-                .compose(StoreUtil.<Parsed, Key>repeatWhenCacheEvicted(refreshSubject, key));
+                .compose(StoreUtil.<Parsed, Key>repeatWhenSubjectEmits(refreshSubject, key));
     }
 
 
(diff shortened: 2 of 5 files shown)


filesystem/src/main/java/com/nytimes/android/external/fs3/SourcePersisterFactory.java
```diff
@@ -78,4 +78,6 @@ private SourcePersisterFactory() {
         return SourcePersister.create(fileSystem);
     }
 
+
+
```
tiwiz: 3 empty lines seem a bit odd?

store/src/main/java/com/nytimes/android/external/store3/base/impl/RealInternalStore.java
```diff
@@ -211,15 +212,21 @@ void notifySubscribers(Parsed data) {
     @Nonnull
     @Override
     public Observable<Parsed> stream(@Nonnull Key key) {
```
tiwiz, line 244: Might the `stream` name that return an `Observable` cause a little bit of confusion with Java 8 `Stream` API?

store/src/main/java/com/nytimes/android/external/store3/base/impl/RealInternalStore.java
```diff
@@ -79,7 +80,7 @@
     public Observable<Parsed> getRefreshing(@Nonnull final Key key) {
         return get(key)
                 .toObservable()
-                .compose(StoreUtil.<Parsed, Key>repeatWhenCacheEvicted(refreshSubject, key));
+                .compose(StoreUtil.<Parsed, Key>repeatWhenSubjectEmits(refreshSubject, key));
```
tiwiz, line 89: We get an `Observable`, we convert it to a `Single` and then back to `Observable`?

store/src/test/java/com/nytimes/android/external/store3/StreamOneKeyTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+        when(persister.write(barCode, TEST_ITEM))
+                .thenReturn(Single.just(true));
+        when(persister.write(barCode, TEST_ITEM2))
+                .thenReturn(Single.just(true));
+    }
+
+    @Test
+    public void testStream() {
+        TestObserver<String> streamObservable = store.stream(barCode).test();
```
tiwiz, line 64: Can we check that it calls the correct methods?

store/src/test/java/com/nytimes/android/external/store3/StreamOneKeyTest.java
```diff
@@ -0,0 +1,85 @@
...
+        //get for another barcode should not trigger a stream for barcode1
+        when(fetcher.fetch(barCode2))
+                .thenReturn(Single.just(TEST_ITEM));
+        when(persister.read(barCode2))
+                .thenReturn(Maybe.empty())
+                .thenReturn(Maybe.just(TEST_ITEM));
+        when(persister.write(barCode2, TEST_ITEM))
+                .thenReturn(Single.just(true));
+        store.get(barCode2).test().awaitCount(1);
+        streamObservable.assertValueCount(2);
```
tiwiz, line 83: Can we check that we are calling the correct methods?

store/src/main/java/com/nytimes/android/external/store3/base/impl/RealInternalStore.java
digitalbuddha: breaking change will leave.

store/src/test/java/com/nytimes/android/external/store3/StreamOneKeyTest.java
digitalbuddha: by asserting the results it will only succeed if proper methods on persister and fetcher were called.

store/src/main/java/com/nytimes/android/external/store3/base/impl/RealInternalStore.java
digitalbuddha: ugh yes this needs to be cleaned up. will do it in another pr good catch!

store/src/main/java/com/nytimes/android/external/store3/base/impl/RealInternalStore.java
tiwiz: Maybe a little bit of documentation could help, tho?

store/src/test/java/com/nytimes/android/external/store3/StreamOneKeyTest.java
tiwiz: But a few assertions more could serve as documentation for the future, so that we don't forget about the flow :-)

ramonaharrison (approved):

PR #264, opened 2017-08-07
Upon calling fetch, notify the refresh subject.

PaulWoitaschek:
To ensure getRefreshing receives the latest data, the refresh mechanism has to be triggered.

### DIFF
--- a/store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
+++ b/store/src/main/java/com/nytimes/android/external/store/base/impl/RealInternalStore.java
@@ -166,7 +166,8 @@ final class RealInternalStore<Raw, Parsed, Key> implements InternalStore<Parsed,
     @Nonnull
     @Override
     public Observable<Parsed> fetch(@Nonnull final Key key) {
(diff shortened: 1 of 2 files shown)


digitalbuddha:
Any reason you can't merge the getRefreshing observable with stream(key).

PaulWoitaschek:
What do you mean by that? Feel free to close this and drop it yourself in the rx2 branch.

digitalbuddha:
There is a public api called store.stream(key) which does exactly that you are after. You could do a store.stream(key).mergeWith(getrefreshing(key))

PaulWoitaschek:
Because I think it would be the correct behavior of `getRefreshing()`. I can't think why you would be interested in making a new call upon clear, but not be notified about data changes.

If you use it, you always want to be up to date with the latest data.

digitalbuddha:
Would you agree that this is a breaking change? I don't see a reason to change behavior if your use case can be currently achieved through above example. I'd prefer to keep apis simple as possible. Feel free to use kotlin locally to create the merged observable above in your project.

PR #280, opened 2017-10-23
#275 feature - added ParsingFetcher that wraps Raw type Parser and Fetcher

magillus:
Reading and parsing network/file fetch can be done at same step with single ParsingFetcher

updated from https://github.com/NYTimes/Store/pull/279

Issue #275, opened 2017-10-07
RealStore with Fetcher<Raw,Key> , Persister<Parsed,Key> and Parser<Raw,Parsed> ?


magillus:
I have use case where the RealStore constructor would be needed:
Current code:
```
    public RealStore(Fetcher<Parsed, Key> fetcher,
                     Persister<Parsed, Key> persister) {
   ...
    public <Raw> RealStore(Fetcher<Raw, Key> fetcher,
                           Persister<Raw, Key> persister,
                           final Parser<Raw, Parsed> parser) {
```
I would like to add:

  ```
  public <Raw> RealStore(Fetcher<Raw, Key> fetcher,
                           Persister<Parsed, Key> parsedPersister,
                           final Parser<Raw, Parsed> parser,
                           final Persister<Raw, Key> responsePersister //optional
                 ) {
```
What I would expect from such store is to fetch from Fetcher `Raw` and parse with `parser` then store with `parsedPersister` also fallback (optional) `responsePersister`  - like file

the `parsedPersister` could be db storage that the requested data would parse and update.

My current solution is first constructor and Fetcher does parsing

**Comment by digitalbuddha** [comment]
That one is tricky since it will break/fork the internal logic of how stores work, currently the data flow is to parse any source coming from network or disk. The reason for this is to allow streaming from network clients to disk.  To add this feature we would need to someone restrict the streaming api.  Since a fetcher's return is an `Observable` I think your best bet is to do what you are doing and use 
``` 
public RealStore(Fetcher<Parsed, Key> fetcher,
                     Persister<Parsed, Key> persister)
 ```

but have your fetcher implementation look something like 
`key->fooApi.makeRequest().map(it->fooParser.parse(it)`

magillus:
Thank you for response, I see the reason for it.
That is what I do right now, my Fetcher wraps network layer fetcher and parser.

```
abstract class BaseFetcher<Parsed, Key>(private val rawFetcher: Fetcher<BufferedSource, Key>,
                                        private val parser: Parser<BufferedSource, Parsed>)
    : Fetcher<Parsed, Key> {
    override fun fetch(key: Key): Single<Parsed> {
        return rawFetcher.fetch(key).subscribeOn(Schedulers.io())
                .map(parser)
    }
}
```

and calling `RealStore<Parsed, Key>(networkFetcher:BaseFetcher<Parsed, Key>, dbPersister: Persister<Parsed, Key>)`

digitalbuddha:
[...]

PR #282, opened 2017-10-30
#275 feature - added ParsingFetcher that wraps Raw type Parser and Fetcher

ramonaharrison:
Reopening #280 by @magillus to run Travis CI.

Reading and parsing network/file fetch can be done at same step with single ParsingFetcher.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -30,6 +30,7 @@ android {
     packagingOptions {
         exclude 'META-INF/rxjava.properties'
     }
+    buildToolsVersion '26.0.2'
 }
 
 dependencies {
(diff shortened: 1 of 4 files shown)


app/build.gradle
```diff
@@ -30,6 +30,7 @@ android {
     packagingOptions {
         exclude 'META-INF/rxjava.properties'
     }
+    buildToolsVersion '26.0.2'
```
magillus, line 33: travis fails because of this version of build tools.
Maybe we can change it or you can update Travis box/agent @ramonaharrison

ramonaharrison: Yes, I've reopened your original PR #280 -- the Travis request seems to be working now. 

I opened #283 to update build tools in the Travis script. As soon as that's merged we should be good to retest your PR.

PR #286, opened 2017-11-14
Changed weird MemoryPolicy to 24 hours

StefMa:
I don't know why it was designed like that but the current implementation looks just strange to me.
Moving 24 hours to seconds and set the TimeUnit to seconds..

Why not just directly use HOURS? 🤔  😄


ramonaharrison (approved):
LGTM... thank you for the catch and small fix! :)

We'll merge after one more reviewer approves.

alexio (approved):

PR #289, opened 2017-11-18
Fix a typo in the documentation

stoyicker:
Also, if I may the chance to ask - I've been using store-kotlin myself for a while now without any issues, nor have others raised any afaik, and it has an acceptable amount of tests imho. What else would be needed to decide that those `@Experimental` annotations can be taken out?


digitalbuddha:
I've actually used it myself and look so good! Feel free to remove the experimental annotation, it doesn't seem like we changed the api once since implementing. Thank you for your hard work and continued help with the project. I love the android community.

PR #301, opened 2017-12-08
[WIP] Expose persister fetcher

### LINKED ISSUE #291: Persister & Fetcher accessable in Store-Interface
Currently we developing some architecture we we put two stores and a fetcher into another class.
Which reads like that:
```kotlin
liveStore: Store,
notLiveStore: Store,
notLivePersister: Persister
```

Which reads a little bit odd because the `notLivePersister` is **already a part** of the `notLiveStore`.

So, what do you think to add 
```kotlin
@Nullable getPersister(): Persister
@Nullable getFetcher(): Fetcher
```
to the Store-Interface? 🤔
Maybe we can also add the Policy and the Parser, but that is not important as the two others IMO 😅

What do you think?

StefMa:
This will fixes #291 

Currently WIP.
Not ready for review.

### DIFF
--- a/store/src/main/java/com/nytimes/android/external/store3/base/InternalStore.java
+++ b/store/src/main/java/com/nytimes/android/external/store3/base/InternalStore.java
@@ -3,11 +3,12 @@ package com.nytimes.android.external.store3.base;
 import com.nytimes.android.external.store3.base.impl.Store;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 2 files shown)


digitalbuddha:
Close or fix?

StefMa:
I have a question in #291 about it which wasn't answered..
I would do it if you can help me. Otherwise you can close this and the issue 👍

PR #315, opened 2018-02-18
Add missing reading of expire-after-policy when creating a NoopPersister

### LINKED ISSUE #312: NoopPersister crashes if expire-after-access memory policy is set
Hello, it appears that `Store` doesn't consider expire-after-access policy when a persister isn't present. The constructor of [NoopPersister(MemoryPolicy)](https://github.com/NYTimes/Store/blob/4109df3a336d6b73d29afdd351b683f388328ae1/store/src/main/java/com/nytimes/android/external/store/util/NoopPersister.java#L26) only includes expire-after-write rule when building its `networkResponses` field, resulting in an exception.

saket:
Fixes #312. I couldn't figure out a way to ensure the memory policies are used while creating cache in `NoopPersister` though.

I also made small style changes. Feel free to ask me to remove them if you want.

### DIFF
--- a/store/src/main/java/com/nytimes/android/external/store3/util/NoopPersister.java
+++ b/store/src/main/java/com/nytimes/android/external/store3/util/NoopPersister.java
@@ -38,20 +38,15 @@ public class NoopPersister<Raw, Key> implements Persister<Raw, Key>, Clearable<K
     }
 
     public static <Raw, Key> NoopPersister<Raw, Key> create(MemoryPolicy memoryPolicy) {
-        //For some reason PMD requires a local variable instead of modifying the passed one.
-        MemoryPolicy memPolicy;
-
         if (memoryPolicy == null) {
-            memPolicy = MemoryPolicy
+            MemoryPolicy defaultPolicy = MemoryPolicy
                 .builder()
                 .setExpireAfterWrite(24)
                 .setExpireAfterTimeUnit(TimeUnit.HOURS)
                 .build();
-        } else {
-            memPolicy = memoryPolicy;
+            return new NoopPersister<>(defaultPolicy);
         }
-
-        return new NoopPersister<>(memPolicy);
+        return new NoopPersister<>(memoryPolicy);
     }
 
     @Nonnull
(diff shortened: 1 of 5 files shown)


ramonaharrison (approved):
This looks good, thank you for submitting. We'll get one more reviewer to take a look + maybe we can help figure out a way to enhance the tests to ensure the memory policies are used.

Otherwise, LGTM. Thanks!!

store/src/main/java/com/nytimes/android/external/store3/util/NoopPersister.java
```diff
@@ -7,45 +7,46 @@
...
+                .expireAfterAccess(memoryPolicy.getExpireAfterAccess(), memoryPolicy.getExpireAfterTimeUnit())
+                .build();
+
+        } else if (memoryPolicy.hasWritePolicy()) {
+            networkResponses = CacheBuilder.newBuilder()
+                .expireAfterWrite(memoryPolicy.getExpireAfterWrite(), memoryPolicy.getExpireAfterTimeUnit())
+                .build();
+        } else {
+            throw new IllegalArgumentException("No expiry policy set on memory-policy.");
+        }
```
cybo42, line 33: Technically `CacheBuilder` lets you set both `expireAfterAccess` and `expireAfterWrite` although it gives preference to `expireAfterAccess`. It's a bit of an edge case, but maybe update this block to handle both being set.

cybo42 (approved):
LGTM

commented on an edge case, but not sure if we need to handle it or not.

store/src/main/java/com/nytimes/android/external/store3/util/NoopPersister.java
saket: Thanks for the review! If both are set, should we throw an exception?

saket:
Hey folks, any updates on this?

ramonaharrison:
I think this looks good to go. I'd like to make a 3.0.1 release available first; will merge once that has completed. Thanks!

Issue #317, opened 2018-03-03
Store for Java?


stoyicker:
I was thinking today about Store and it being an Android library. That's cool and all, but upon a superficial search on GitHub there doesn't seem to be a reason to it - its important dependencies are RxJava, Retrofit, Moshi, Gson and Jackson, none of which are Android-specific, and I didn't find any android.* imports. Only thing that would seem to be required as far as I can see is a bit of a refactoring regarding things like removing android manifests, and I'd be happy to do it. What do you think of this? Is it something you'd be open to?

ramonaharrison:
I don't see any reason not to do this -- it would be great to make Store available for a wider set of uses.

If this is something you're interested in doing, we'd welcome a PR.

stoyicker:
Thanks, I'll see what I can get together this weekend.

PR #319, opened 2018-03-10
Remove unnecessary Android-specific stuff

stoyicker:
Begins to address #317 by removing Android manifests which are not doing anything and some unrequired Gradle stuff. Also some Android references in the README have been removed.

However:
* CI runs the check task on the entire project, which means Android is needed for the sample app to work. My recommendation would be that we change the sample to be raw Java, which would allow skipping Android on CI, and therefore make it faster. On top of that, if the artifact is to be changed to be non-Android-specific, it's probably a good idea to not to have the sample not be Android-y anyway. 
* The package name for the most of the code, as well as the maven group id, includes "android" in the name. Both of these should IMHO be refactored out, but I didn't want to take this decision on my own.
* The "commercial effort" (changing Android to Java in things like the repo short-description and tags, and whatever else, is still left to do as that can't be done with a PR


digitalbuddha:
This seems like the first pr that should work towards next major version  of store. We can change the maven coordinates and remove all the android references. I agree about changing sample project to java.

digitalbuddha:
Give me a bit to get a release plan together and we will merge this in. Thank you for contribution!

digitalbuddha:
could someone on nyt side merge pls

tiwiz (approved):

tiwiz:
@digitalbuddha I would wait for one more approval before merging it in

ramonaharrison (approved):

PR #322, opened 2018-03-21
[WIP] File system erase ability

bishiboosh:
I've added the clear functionality to the default file system persisters.

As the clear operation can take some times, I've also changed `clear` signature to be `Completable`

### DIFF
--- a/store/src/main/java/com/nytimes/android/external/store3/base/impl/RealInternalStore.java
+++ b/store/src/main/java/com/nytimes/android/external/store3/base/impl/RealInternalStore.java
@@ -7,11 +7,16 @@ import com.nytimes.android.external.store3.base.Fetcher;
 import com.nytimes.android.external.store3.base.InternalStore;
 import com.nytimes.android.external.store3.base.Persister;
 import com.nytimes.android.external.store3.util.KeyParser;
+
 import java.util.AbstractMap;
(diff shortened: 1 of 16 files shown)


store/src/main/java/com/nytimes/android/external/store3/base/impl/RealInternalStore.java
```diff
@@ -281,23 +286,30 @@ public void clearMemory() {
...
     @Override
-    public void clear() {
-        for (Key cachedKey : memCache.asMap().keySet()) {
-            clear(cachedKey);
-        }
+    public Completable clear() {
+        return Completable.concat(memCache
+                .asMap()
+                .keySet()
+                .stream()
```
digitalbuddha, line 298: Isn't this java8 stream api? Won't work on android.

digitalbuddha:
Will hold off to merging until next major store version which we are starting to prepare. This is due to breaking change with current functionality.

bishiboosh:
@digitalbuddha sure. I've detected some problems with my code too, so I'll wait until the next version and come back to clean this then

digitalbuddha:
howdy ho, will this be updated or should I close?

bishiboosh:
@digitalbuddha I don't think I have the time available right now, close it and I'll take another stab at it later :)

PR #338, opened 2018-05-25
StoreRoom

### LINKED ISSUE #215: Spike: replace file system with Room/sql
Now that android has a proper orm, we should Spike on depreciating our filesystem and moving to Room (new orm). It seems like you can save nested objects without having to write complex logic yourself, that was always the blocker imo to using sql.  

Ideally store should be able to work with retrofit returning a pojo which is the written to and read from Room. 

Alternatively we can create Room persisters that work with sql while keeping current File System stuff the same. 

Yay exciting times in android!

digitalbuddha:
closes #215 

This PR is a working implementation of a Store that can use a RoomPersister.  While not in its final form I wanted to open the pr to get feedback.  I'll annotate the important parts.

Example usage (will move to readme)
example usage:
```kotlin
@Entity
data class User(
        @PrimaryKey(autoGenerate = true)
        var uid: Int = 0,
        val name: String)

@Dao
interface UserDao {
    @Query("SELECT name FROM user")
    fun loadAll(): Flowable<List<String>>

    @Insert
    fun insertAll(vararg users: User)

}

@Database(entities = arrayOf(User::class), version = 1)
abstract class AppDatabase : RoomDatabase() {
    abstract fun userDao(): UserDao
}

val db = Room.databaseBuilder(SampleApp.appContext!!, AppDatabase::class.java, "db").build()


val store = RoomInternalStore(
        Fetcher<User, String> { Single.just(User(name = "Mike")) },
        object : RoomPersister<User, List<String>, String> {

            override fun read(key: String): Observable<List<String>> {
                return db.userDao().loadAll().toObservable()
            }

            override fun write(key: String, user: User) {
                db.userDao().insertAll(user)
            } 
        })
```


app/src/main/java/com/nytimes/android/sample/SampleApp.kt
```diff
@@ -0,0 +1,137 @@
...
+        initPersister();
+        nonPersistedStore = provideRedditStore();
+        persistedStore=providePersistedRedditStore();
+        RoomSample()
+
+
+    }
+
+    private fun RoomSample() {
+        var foo = store.get("")
```
digitalbuddha: see SampleRoomStore.kt for the implementation of this store, biggest takeaway is the ability to treat each store subscription as an endless stream backed by a Room Query. Now anytime you subscribe to `get` if someone calls `store.fresh` or if another `get` call busts through the cache, we will emit to all of the previous `get/fresh` subscribers. Store becomes truly reactive when backed by room

app/src/main/java/com/nytimes/android/sample/SampleRoomStore.kt
```diff
@@ -0,0 +1,74 @@
...
+import com.nytimes.android.external.store3.base.impl.StalePolicy
+import com.nytimes.android.external.store3.base.impl.room.RoomInternalStore
+import com.nytimes.android.external.store3.base.room.RoomPersister
+import io.reactivex.Flowable
+import io.reactivex.Observable
+import io.reactivex.Single
+
+
+// File: User.java
+@Entity
```
digitalbuddha: Simple implementation of a RoomDB as a proof of concept

app/src/main/java/com/nytimes/android/sample/SampleRoomStore.kt
```diff
@@ -0,0 +1,74 @@
...
+
+// File: AppDatabase.java
+@Database(entities = arrayOf(User::class), version = 1)
+abstract class AppDatabase : RoomDatabase() {
+    abstract fun userDao(): UserDao
+}
+
+val db = Room.databaseBuilder(SampleApp.appContext!!, AppDatabase::class.java, "db").build()
+
+val persister = object : RoomPersister<User, List<String>, String>() {
```
digitalbuddha: A new persister needed to be created that is exposed as an observable rather than a single.  The RoomPersister takes a third param as you can have the freedom to have different queries for reading and writing. In this example writing is always done by adding a single user to the DB while read returns the list of all users. Another use case is writing a User but only reading a part of the user

app/src/main/java/com/nytimes/android/sample/SampleRoomStore.kt
```diff
@@ -0,0 +1,74 @@
...
+    override fun write(key: String, user: User) {
+        db.userDao().insertAll(user)
+    }
+
+}
+
+
+
+//store
+class SampleRoomStore(fetcher: Fetcher<User, String>,
```
digitalbuddha: Nothing crazy here besides a few interface changes that lead to a RoomInternalStore.  TODO: create builders for instantiation

store/src/main/java/com/nytimes/android/external/store3/base/BasePersister.java
```diff
@@ -0,0 +1,8 @@
+package com.nytimes.android.external.store3.base;
+
+
+
+
+public interface BasePersister {
```
digitalbuddha, line 6: Since we now have `RoomPersister` and `Persister` I made a marker interface for use with `StoreUtil` for things like checking stale policy or clearing the persister

store/src/main/java/com/nytimes/android/external/store3/base/impl/CacheFactory.java
```diff
@@ -58,4 +59,50 @@ private CacheFactory() {
                     .build();
         }
     }
+
+    public static <Key, Parsed> Cache<Key, Observable<Parsed>> createRoomCache(MemoryPolicy memoryPolicy) {
```
digitalbuddha: needed a new cache implementation that holds observables rather than singles to keep the "live query" setup

store/src/main/java/com/nytimes/android/external/store3/base/impl/room/RoomInternalStore.java
```diff
@@ -0,0 +1,216 @@
...
+import io.reactivex.disposables.Disposable;
+
+/**
+ * Store to be used for loading an object from different data sources
+ *
+ * @param <Raw>    data type before parsing, usually a String, Reader or BufferedSource
+ * @param <Parsed> data type after parsing
+ *                 <p>
+ */
+public class RoomInternalStore<Raw, Parsed, Key> implements RoomStore<Parsed, Key> {
```
digitalbuddha: RoomInternalStore is the meat of it.  Notice that we don't need `stream` or `getRefreshing` as both of those can be achieved by Room without the need of the helpers.  Update a query that a RoomPersister is listening to and `get` will refresh on its own. If not preferable users can transform the `get` back into a `single`

digitalbuddha: Most logic in this class stayed the same but changed to use Observable rather than Single

store/src/main/java/com/nytimes/android/external/store3/base/room/RoomPersister.java
```diff
@@ -0,0 +1,33 @@
...
+
+import io.reactivex.Observable;
+
+/**
+ * Interface for fetching data from persister
+ * when implementing also think about implementing PathResolver to ease in creating primary keys
+ *
+ * @param <Raw> data type before parsing
+ */
+public abstract class RoomPersister<Raw, Parsed, Key> implements RoomDiskRead<Parsed, Key>, RoomDiskWrite<Raw, Key>, BasePersister {
```
digitalbuddha: Parsers are unnecessary since the DB read query can be a "parsed" version of whatever the written "raw" data is

tiwiz (approved):

digitalbuddha:
I'll migrate tests as well to make sure the Room impl works as well as the regular one.  I'll also mark this implementation as experimental

app/build.gradle
```diff
@@ -1,6 +1,7 @@
 apply plugin: 'com.android.application'
+apply plugin: 'kotlin-android'
 apply plugin: 'com.getkeepsafe.dexcount'
-
+apply plugin: 'kotlin-kapt'
```
brianPlummer, line 4: woohoo!

buildsystem/dependencies.gradle
```diff
@@ -15,13 +15,13 @@ allprojects {
...
-        targetSdk            : 25,
-        compileSdk           : 25,
+        targetSdk            : 27,
+        compileSdk           : 27,
         buildTools           : '27.0.3',
         kotlin               : '1.1.2-5',
 
         // UI libs.
-        supportLibs          : '25.1.1',
+        supportLibs          : '26.1.0',
```
brianPlummer: should be 27 as well...

app/src/main/java/com/nytimes/android/sample/SampleApp.kt
```diff
@@ -0,0 +1,137 @@
...
+        return GsonBuilder()
+                .registerTypeAdapterFactory(GsonAdaptersModel())
+                .create()
+    }
+
+    companion object {
+
+        var appContext: Context? = null
+    }
+}
```
brianPlummer: format and remove EL for this file..

app/build.gradle
```diff
@@ -48,9 +50,25 @@ dependencies {
     annotationProcessor libraries.immutablesValue // <-- for annotation processor
     compileOnly libraries.immutablesValue // <-- for annotation API
     compileOnly libraries.immutablesGson // for annotations
-    implementation 'com.nytimes.android:store3:3.0.1'
-    implementation 'com.nytimes.android:cache3:3.0.1'
-    implementation 'com.nytimes.android:middleware3:3.0.1'
-    implementation 'com.nytimes.android:filesystem3:3.0.1'
+    implementation project(':store')
```
brianPlummer, line 53: might need to flip before checking in....

brianPlummer:
LGTM besides the nitpicks :)

app/build.gradle
digitalbuddha: will do once I can push the new snapshot thanks!

digitalbuddha:
Added prelimanary tests, marked RoomStore as Experimental and added tests, bump rev up to 3.1.0

Pr is ready to merge IMO.

Follow up from me will be a blog post, more tests and a cleanup of the example that is currently living in `SampleApp`

@tiwiz wake up and review my code :-D

digitalbuddha:
example usage:
```

@Entity
data class User(
        @PrimaryKey(autoGenerate = true)
        var uid: Int = 0,
        val name: String)

@Dao
interface UserDao {
    @Query("SELECT name FROM user")
    fun loadAll(): Flowable<List<String>>

    @Insert
    fun insertAll(vararg users: User)

}

@Database(entities = arrayOf(User::class), version = 1)
abstract class AppDatabase : RoomDatabase() {
    abstract fun userDao(): UserDao
}

val db = Room.databaseBuilder(SampleApp.appContext!!, AppDatabase::class.java, "db").build()


val store = RoomInternalStore(
        Fetcher<User, String> { Single.just(User(name = "Mike")) },
        object : RoomPersister<User, List<String>, String> {

            override fun read(key: String): Observable<List<String>> {
                return db.userDao().loadAll().toObservable()
            }

            override fun write(key: String, user: User) {
                db.userDao().insertAll(user)
            } 
        })
```

tiwiz (approved):

wdziemia (approved):

app/src/main/java/com/nytimes/android/sample/SampleApp.kt
```diff
@@ -0,0 +1,133 @@
...
+        var foo = store.get("")
+                .subscribeOn(Schedulers.io())
+                .observeOn(AndroidSchedulers.mainThread())
+                .subscribe({ strings1 -> val success = strings1 != null }) { throwable -> throwable.stackTrace }
+
+        foo = Observable.timer(15, TimeUnit.SECONDS)
+                .subscribe { again() }
+    }
+
+    private fun again() {
```
alexio: Change to more descriptive name?

app/src/main/java/com/nytimes/android/sample/SampleRoomStore.kt
```diff
@@ -0,0 +1,56 @@
...
+    fun insertAll(user: User)
+
+}
+
+@Database(entities = arrayOf(User::class), version = 1)
+abstract class AppDatabase : RoomDatabase() {
+    abstract fun userDao(): UserDao
+}
+
+val db = Room.databaseBuilder(SampleApp.appContext!!, AppDatabase::class.java, "db").build()
```
alexio: Reimplement to avoid using the `!!` operator? 
Can be as simple as creating a builder that just takes a context.

app/src/main/java/com/nytimes/android/sample/SampleApp.kt
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    internal fun provideGson(): Gson {
+        return GsonBuilder()
+                .registerTypeAdapterFactory(GsonAdaptersModel())
+                .create()
+    }
+
+    companion object {
+        var appContext: Context? = null
```
alexio, line 131: Not great, lets avoid having objects create a static reference to themselves?

store/src/main/java/com/nytimes/android/external/store3/base/impl/CacheFactory.java
alexio: You can keep it simple and have a method level generic for the singles/maybes/observables.

`private static <Key, Value> Cache<Key, Value> createBaseCache(MemoryPolicy memoryPolicy)`

alexio (changes requested):
Some code quality request changes.

alexio (approved):

app/build.gradle
```diff
@@ -48,9 +50,25 @@ dependencies {
...
+//    androidTestImplementation "android.arch.persistence.room:testing:$room_version"
+    // optional - RxJava support for Room
+    implementation "android.arch.persistence.room:rxjava2:$room_version"
 }
+repositories {
+    mavenCentral()
+}
+
+configurations.all {
+    resolutionStrategy.force 'com.android.support:support-v4:26.1.0'
```
trietbui85, line 73: It seems we're still using support lib v26.1.0 while there is latest one v27.1.1. Any plan to upgrade?

PR #349, opened 2018-07-07
convert moshi middleware to kotlin

brianPlummer:
This converts the moshi middleware to kotlin


brianPlummer:
@digitalbuddha

middleware-moshi/src/test/java/com/nytimes/android/external/store3/middleware/moshi/MoshiSourceParserTest.kt
```diff
@@ -0,0 +1,98 @@
...
+import java.nio.charset.Charset
+
+class MoshiSourceParserTest {
+    @Rule
+    @JvmField
+    var expectedException = ExpectedException.none()
+    @Mock
+    internal var fetcher: Fetcher<BufferedSource, BarCode>? = null
+    @Mock
+    internal var persister: Persister<BufferedSource, BarCode>? = null
```
ramonaharrison: can we make the mocks `lateinit` here and in other test classes? that way we won't need null assertions whenever they are used

goldstei (approved):

middleware-moshi/src/test/java/com/nytimes/android/external/store3/middleware/moshi/MoshiSourceParserTest.kt
brianPlummer: done!

PR #373, opened 2018-12-20
Create InMemoryFileSystem for testing purposes

tiwiz:


digitalbuddha (changes requested):
Without tests I can't tell if you're code works correctly :-P

digitalbuddha:
ping to add tests. Y'all still write tests right?

PR #379, opened 2019-02-11
[DO NOT MERGE] Feature/Coroutines

digitalbuddha:
Store module is done, converted about 20% of tests.

TODO: need to do some conversion so that middleware works with suspend instead of observables

Last step will be to change the package to store4 which will be 100% kotlin


fabioCollini:
The `Fetcher` is defined as a Kotlin interface but it's quite awkward to implement it:

    return StoreBuilder.barcode<RedditData>()
            .fetcher(object : Fetcher<RedditData, BarCode> {
                override suspend fun fetch(key: BarCode): RedditData {
                    return provideRetrofit().fetchSubreddit(key.key, "10").await()
                }
            })

Should we define it as a `typealias` instead? Something like:

`typealias Fetcher<Raw, Key> = suspend (Key) -> Raw`

In this way the code can be simplified to:

    return StoreBuilder.barcode<RedditData>()
            .fetcher { key -> provideRetrofit().fetchSubreddit(key.key, "10").await() }

digitalbuddha:
Yes that's amazing

fabioCollini:
I am looking at `RealInternalStore` and I have a couple of doubts:

- the `inFlightScope` and `memoryScope` are defined as `CoroutineScope(SupervisorJob())`, in this way I think that the errors are not managed. In case of `inFlightScope` probably `GlobalScope` with a `try/finally` can be used. If it's never cancelled `GlobalScope` can work fine
- all the concurrency is managed in this class so I am not sure it's not necessary to define all the invoked methods as `suspend`. For example the `readDisk` is defined as `suspend` even if it's not necessary because it doesn't invoke any real `suspend` methods (more info here https://medium.com/@elizarov/blocking-threads-suspending-coroutines-d33e11bf4761). So the `Persister` defined as `suspend` can be useful to support custom persister that use coroutines but are not useful for the current implementations

Issue #381, opened 2019-03-06
fetchWithResult returns NETWORK even on network errors


fabioCollini:
The `fetchWithResult` method always returns a `Result` created using `createFromNetwork`:

```
public Single<Result<Parsed>> fetchWithResult(@Nonnull Key key) {
    return fetch(key).map(Result::createFromNetwork);
}
```

The problem is that if there is an error the `fetch` method loads the data from the cache but the result is created using `createFromNetwork`.
I can work to fix it but unfortunately it doesn’t seem easy to fix, the only solution I can find is to refactor the `fetch` and  `fetchWithResult` in the `RealInternalStore` class. The `fetch` method should call `fetchWithResult` and execute a map to return just the data. What do you think?

digitalbuddha:
close?

fabioCollini:
Yes!

PR #382, opened 2019-03-06
Fix fetchWithResult return value in case of fetch error

fabioCollini:
This PR fixes the issue #381, it contains a refactoring to return the `Result` on `fetchWithResult` instead of returning always `NETWORK` as source


store/src/test/java/com/nytimes/android/external/store3/StoreTest.java
```diff
@@ -99,6 +101,30 @@ public void testSimpleWithResult() {
         verify(fetcher, times(1)).fetch(barCode);
     }
 
+    @Test
+    public void testFetchWithResultReturnsCacheOnErrors() {
```
digitalbuddha, line 105: 💯

digitalbuddha (approved):

ramonaharrison (approved):

